Issue #4, opened 2016-07-09
Unable to get it working


mickaelperrin:
Hi,

Thanks for working on this project.

I tried to configure my own docker project ith docker_sync and also tried the boiler plate but each ends with the error:

          ok  Starting unison
    /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_strategy/unison.rb:38:in `sync': undefined method `join' for nil:NilClass (NoMethodError)
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_strategy/unison.rb:93:in `start_container'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_strategy/unison.rb:32:in `run'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_process.rb:64:in `run'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_manager.rb:103:in `block in run'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_manager.rb:102:in `each'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_manager.rb:102:in `run'
	from /Users/mickaelperrin/tmp/docker_sync/tasks/sync.thor:32:in `start'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
	from ../../docker_sync/bin/docker-sync:11:in `<main>'

It looks like the `args` are not properly parsed. 

Am I missing something ?

Regards,

EugenMayer:
thanks for fixing the bug yourself - pitty that slipped through my test - i do not have any unit tests yet ( not easy to write it for this tool though )

Issue #5, opened 2016-07-09
Started, but not syncing unless on restart of docker-sync


iwaffles:
I've managed to get docker-sync running and it seems to share the volume data when I initially run `docker-sync start`, however, after editing anything in the directory that it's watching it doesn't appear to sync any data.

Restarting docker-sync shows a new set of data, but that defeats the purpose of running it once and letting data sync when saved/updated.

Is there possibly a step I'm missing here? I don't have any watch excludes setup and I'm using rsync 3.1.2 on OS X. I don't see any errors in the logs, either.

iwaffles:
For what it's worth I tried running the `fswatch` command with thor manually and it seems to return `Could not find command "sync:sync".`

EugenMayer:
@iwaffles thats for sure not the desired behavior. Which gem version are you using? looks like you are below < 0.0.7? if yes, please upgrae.

Could you paste your docker-sync.yml?

iwaffles:
Ah yes it appears that I'm on 0.0.5. Upgraded to 0.0.8 and all seems to be working!

Great work @EugenMayer!

iwaffles:
For what it's worth, here's my docker-sync file:

```yaml
options:
  # verbose: true
syncs:
  appcode:
    src: '.'
    dest: '/app'
    sync_host_ip: 'localhost'
    sync_host_port: 9999
    sync_strategy: 'rsync'
```

PR #6, opened 2016-07-09
[BUGFIX #4] Start fails with error

mickaelperrin:
- Error: lib/docker_sync/sync_strategy/unison.rb:38:in `sync': undefined
  method `join' for nil:NilClass (NoMethodError)
- Fix: looks like sync_options return nothing


mickaelperrin:
Bad issue attached, this is #4 not #5

Issue #9, opened 2016-07-09
docker_sync fails to start with verbose option


mickaelperrin:
I am experiencing sync issues with unison, I try to understand what's going on.

I activated the `verbose` option  and got this issue. It looks like there is a color definition missing.

     /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/shell/color.rb:90:in `const_get': uninitialized constant OK (NameError)
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/shell/color.rb:90:in `set_color'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/shell/basic.rb:98:in `say_status'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/shell.rb:59:in `say_status'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_strategy/unison.rb:83:in `start_container'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_strategy/unison.rb:32:in `run'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_process.rb:64:in `run'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_manager.rb:103:in `block in run'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_manager.rb:102:in `each'
	from /Users/mickaelperrin/tmp/docker_sync/lib/docker_sync/sync_manager.rb:102:in `run'
	from /Users/mickaelperrin/tmp/docker_sync/tasks/sync.thor:32:in `start'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
[...]

PR #10, opened 2016-07-09
[BUGFIX #9] fails to start with verbose option

mickaelperrin:
:ok color is not defined, use :white instead


EugenMayer:
great, nice to have somebody doing a second read. btw. @mickaelperrin intersing docker-sync-mac + lsyncd projects - impressive.

Not sure how lsyncd is superior over unison or rsyncd ( well this one is uni-directonal ) but feel free to implement it as a sync-strategy for docker-sync similar to rsync / lsyncd. Seems like you go the same route with containers with named volumes

Issue #11, opened 2016-07-09
fails to log output with verbose option


mickaelperrin:
/Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/lib/docker_sync/sync_strategy/unison.rb:49:in `sync': undefined local variable or method `out' for #<Docker_Sync::SyncStrategy::Unison:0x007fc942119b18> (NameError)
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/lib/docker_sync/sync_strategy/unison.rb:94:in `start_container'
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/lib/docker_sync/sync_strategy/unison.rb:32:in `run'
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/lib/docker_sync/sync_process.rb:64:in `run'
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/lib/docker_sync/sync_manager.rb:103:in `block in run'
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/lib/docker_sync/sync_manager.rb:102:in `each'
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/lib/docker_sync/sync_manager.rb:102:in `run'
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/tasks/sync.thor:32:in `start'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /Users/mickaelperrin/.rvm/gems/ruby-2.2.1/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
	from /Users/mickaelperrin/tmp/docker-sync-perso/docker_sync/bin/docker-sync:11:in `<main>'

EugenMayer:
i guess you fixed this?

mickaelperrin:
yes thanks for the rapid merge

PR #12, opened 2016-07-09
Issue #11

mickaelperrin:
It looks like `out` is not defined.

Didn't check if that kind of error exist elsewhere in code.


EugenMayer:
This is the single point were this is used this way, it was changed recently

Issue #14, opened 2016-07-09
[FEATURE] More verbosity to unison output


mickaelperrin:
When verbose is activated, `unison`should be more verbose.

PR is incoming.

EugenMayer:
thank you!

EugenMayer:
@mickaelperrin the features your worked on and my update-checker are not released with 0.0.9

PR #15, opened 2016-07-09
[FEATURE #14] Adds more verbosity to unison

mickaelperrin:
When docker_sync is run with verbose option, unison should run also with the verbose flag.


EugenMayer:
sounds fair.

Be aware, you can set any kind of sync options yourself, sync_args: '--debug verbose' in the docker-sync.yml

Still, this is a great contribution, thank you!

PR #18, opened 2016-07-10
[FEATURE] Unison excludes

mickaelperrin:
Example usage:

    sync_excludes: [ '.git', '.idea', 'node_modules' ]


EugenMayer:
Great work - i looked in the docs but could not find it shortly, so i skipped it.

Great you actually doing so much work on unison, since my focus is on rsync a bit more. Nice addition

mickaelperrin:
It looks like I forgot a warning message in the patch. I will add it lately.

PR #19, opened 2016-07-10
[FEATURE] unison > sync uid/gid numerically

mickaelperrin:
Currently all unison sync is done under `root`.

This is an issue if the mounted folder is intended to be used with a non root user, like `www-data`.

This patch ensures that files are sync with the uid/gid of the host user, so that in container we can simply done a `usermod -u $(stat -c '%u' /mountedfolder) www-data`.

I don't know if it should be `default` comportment or if it needs a special flag.

### DIFF
--- a/lib/docker_sync/sync_strategy/unison.rb
+++ b/lib/docker_sync/sync_strategy/unison.rb
@@ -62,6 +62,9 @@ module Docker_Sync
         args.push(@options['src'])
         args.push('-auto')
         args.push('-batch')
+        args.push('-owner')
+        args.push('-group')
+        args.push('-numericids')
         args.push(@options['sync_args']) if @options.key?('sync_args')
         args.push("socket://#{@options['sync_host_ip']}:#{@options['sync_host_port']}/")
         args.push('-debug verbose') if @options['verbose']


mickaelperrin:
In fact, currently `/mountedfolder` is still owned by `root`. Files inside of it are well owned, but it causes problem with directory creation directly in /mountedfolder by non `root` process.

I think I can manage it with a kind of `docker-entrypoint.sh` script in my container, but I would prefer a solution managed directly by `docker_sync`. If you have some ideas...

EugenMayer:
I do not really want to go this route here. Using the UID/GID of the host user is not helping at all, thats why i introduced user/uid + group/gid selections for fsync, see 

https://github.com/EugenMayer/docker_sync/blob/master/example/docker-sync.yml#L27

I can use simple ENV vars here https://github.com/EugenMayer/docker_sync/blob/master/lib/docker_sync/sync_strategy/rsync.rb#L89

since the rsync image does then create this use on the image, then remaps those files on the sync-container, which is then mounted by the app-containers. This way you can setup proper permissions for the destinations container, not just "some arbitrary uid/gid" your host system has, but surely your app-container will not have.

So in the rsync image, we have this entrypoint:

```
OWNER=${OWNER:-nobody}
GROUP=${GROUP:-nogroup}
OWNERID=${OWNERID:-1000}
GROUPID=${GROUPID:-1000}

id -u $GROUP &>/dev/null || (groupadd -g $GROUPID $GROUP && echo "created group $GROUP with id $GROUPID")
id -u $OWNER &>/dev/null || (useradd -N -G $GROUP -u $OWNERID $OWNER --home $VOLUME && echo "created user $OWNER with id $OWNERID")

mkdir -p $VOLUME

chown "${OWNER}:${GROUP}" "${VOLUME}"

[ -f /etc/rsyncd.conf ] || cat <<EOF > /etc/rsyncd.conf
uid = ${OWNER}
gid = ${GROUP}
use chroot = yes
log file = /dev/stdout
reverse lookup = no
[volume]
    hosts deny = *
    hosts allow = ${ALLOW}
    read only = false
    path = ${VOLUME}
    comment = docker volume
EOF
```

Since i cannot see a proper implementation for unison right now, i did not implement it: https://github.com/EugenMayer/docker_sync/blob/master/lib/docker_sync/sync_strategy/unison.rb#L69

I think this will be one of the bigger downsides of unison in general. Two way syncing means also, two way user mapping, and that is not solved at all.

When someone needs proper user mapping, he should rather use rsync, thats just better in this case

mickaelperrin:
Ok, I understand that it shouldn't be the default behaviour, but please consider the PR #21 that adds a special flag to enable that kind of feature.

EugenMayer:
looks like this is a dupe, since changes from this PR are part of the other. Closing this one as dupe then

PR #21, opened 2016-07-10
Unison > option to map uid from host

mickaelperrin:
This pull request adds a special value `from_host` for `sync_userid` and `sync_groupid` options.

When used with unison, it enables syncing of host `uid` `gid` alongside the files.

Additionally it sets the adequate ownership for the mounted folder.

### DIFF
--- a/lib/docker_sync/sync_strategy/unison.rb
+++ b/lib/docker_sync/sync_strategy/unison.rb
@@ -55,9 +55,8 @@ module Docker_Sync
         args = []
 
         unless @options['sync_excludes'].nil?
-          # TODO: does unison support excludes as a command parameter? seems to be a config-value only
           say_status 'warning','Excludes are yet not implemented for unison!', :yellow
-          #  args = @options['sync_excludes'].map { |pattern| "--exclude='#{pattern}'" } + args
+          args = @options['sync_excludes'].map { |pattern| "-ignore='Path #{pattern}'" } + args
         end
         args.push(@options['src'])
         args.push('-auto')
--- a/lib/docker_sync/sync_strategy/unison.rb
+++ b/lib/docker_sync/sync_strategy/unison.rb
@@ -61,13 +61,13 @@ module Docker_Sync
         args.push(@options['src'])
         args.push('-auto')
         args.push('-batch')
-        args.push('-owner')
-        args.push('-group')
-        args.push('-numericids')
+        args.push('-owner') if @options['sync_userid'] == 'from_host'
+        args.push('-group') if @options['sync_groupid'] == 'from_host'
+        args.push('-numericids') if @options['sync_userid'] == 'from_host' || @options['sync_groupid'] == 'from_host'
         args.push(@options['sync_args']) if @options.key?('sync_args')
         args.push("socket://#{@options['sync_host_ip']}:#{@options['sync_host_port']}/")
         args.push('-debug verbose') if @options['verbose']
-        if @options.key?('sync_user') || @options.key?('sync_group') || @options.key?('sync_groupid') || @options.key?('sync_userid')
+        if @options.key?('sync_user') || @options.key?('sync_group') || @options.key?('sync_groupid') && @options['sync_groupid'] != 'from_host' || @options.key?('sync_userid') && @options['sync_userid'] != 'from_host'
           raise('Unison does not support sync_user, sync_group, sync_groupid or sync_userid - please use rsync if you need that')
         end
         return args
@@ -85,6 +85,7 @@ module Docker_Sync
           if exists == ''
             say_status 'ok', "creating #{container_name} container", :white if @options['verbose']
             cmd = "docker run -p '#{@options['sync_host_port']}:#{UNISON_CONTAINER_PORT}' -v #{volume_name}:#{@options['dest']} -e UNISON_VERSION=#{UNISON_VERSION} -e UNISON_WORKING_DIR=#{@options['dest']} --name #{container_name} -d #{UNISON_IMAGE}"
+            cmd = cmd + " && docker exec #{container_name} chown #{Process.uid} #{@options['dest']}" if @options['sync_userid'] == 'from_host'
           else
             say_status 'ok', "starting #{container_name} container", :white if @options['verbose']
             cmd = "docker start #{container_name}"


EugenMayer:
This will only work for the first sync, since your docker exec is not executed when docker-sync sync is executed, after starting the sync, if you create a new file, you will see it having the wrong ownershipt in the container. This needs a different solution.

Also you are using Process.uid - this uid might simply not exist on the unison container and will change - you do create this user pre container start.

I also do not want to run arbitrary docker exec commands which actually should have been things to be part of the entrypoint shell wrapper. We should

a) offer user/uid group/guid as ENV vars to the unison image
b) the unison image should be customised in terms of the entrypoint and create the users the same way we do for rsync
c) we need to ensure that permissions are applied for every new synced file, so this being part either of unison or the "sync" command as a custom docker exec (other solutions preferred).

I am sorry to not offer you a quick-shot solution here, but this needs tender care before we introduce this.

Thanks for your thoughs and the contribution

lib/docker_sync/sync_strategy/unison.rb
```diff
@@ -82,6 +85,7 @@ def start_container
           if exists == ''
             say_status 'ok', "creating #{container_name} container", :white if @options['verbose']
             cmd = "docker run -p '#{@options['sync_host_port']}:#{UNISON_CONTAINER_PORT}' -v #{volume_name}:#{@options['dest']} -e UNISON_VERSION=#{UNISON_VERSION} -e UNISON_WORKING_DIR=#{@options['dest']} --name #{container_name} -d #{UNISON_IMAGE}"
+            cmd = cmd + " && docker exec #{container_name} chown #{Process.uid} #{@options['dest']}" if @options['sync_userid'] == 'from_host'
```
EugenMayer: should not happen here, otherwise its on container start only

lib/docker_sync/sync_strategy/unison.rb
```diff
@@ -61,10 +61,13 @@ def sync_options
...
         args.push('-auto')
         args.push('-batch')
+        args.push('-owner') if @options['sync_userid'] == 'from_host'
+        args.push('-group') if @options['sync_groupid'] == 'from_host'
+        args.push('-numericids') if @options['sync_userid'] == 'from_host' || @options['sync_groupid'] == 'from_host'
         args.push(@options['sync_args']) if @options.key?('sync_args')
         args.push("socket://#{@options['sync_host_ip']}:#{@options['sync_host_port']}/")
         args.push('-debug verbose') if @options['verbose']
-        if @options.key?('sync_user') || @options.key?('sync_group') || @options.key?('sync_groupid') || @options.key?('sync_userid')
+        if @options.key?('sync_user') || @options.key?('sync_group') || @options.key?('sync_groupid') && @options['sync_groupid'] != 'from_host' || @options.key?('sync_userid') && @options['sync_userid'] != 'from_host'
```
EugenMayer: that looks pretty good. Still ,you need to add this option to example/docker-sync.yml and document it. We should then either offer the same for rsync or state that this is unison only and catch this in rsync as a exception the same way i did this for excludes / users in unison before

lib/docker_sync/sync_strategy/unison.rb
mickaelperrin: Don't understand why do we need it elsewhere than docker start ? unison manage uid/gid syncing for other folders, those that will be updated.

EugenMayer: when the dest folder on the unison container has uid 10 and you start the container, sync and then run the chown, the files are properly owned.

When you now create a file on the host FS, a new one, which then triggers fswatch, which triggers docker-sync sync - this will just run unison on the host, without the docker exec. This means, the files have the wrong permissions on the host - or are they handled "magical" stickiness?

mickaelperrin:
Indeed, I intend to work on a new docker image for unison, but was mainly to manage fsevents on the container side.

I don't understand why we have to create users in the volume container. I thought that what's important is uid/gid not users. Depending on where the folder should be mounted you could use the same uid but mapped to a different user.

For example: 
- `src` is mounted in nginx you could want to map the `uid` to the `nginx` user.
- `src` is mounted in php-fpm, you could want to map the `uid` to the `www-data` user.

In my opinion, user creation / update should be done in the app / stack side not in the volume container.

EugenMayer:
Sure, the app-container does need this user and needs to create it - not doubt.

But since the unison as the rsync container do work as a kind of "filesystem proxy", AFAIU you need the target user there, since you then mount the volume of the sync-container and that exact uid will be mapped on the target app contaner.

For the fsevent part, did you find https://github.com/codekitchen/fsevents_to_vm ?

mickaelperrin:
For what I know, `fsevents_to_vm` was done to forward `fs`events on NFS mounted folder in the `boot2docker` vm.

Before your project, I was doing a kind of two-way sync with two rsync and two fsevents watcher, one on the host, the other on the container. Worked, but really heavy CPU when syncing vendor / core Drupal after bootstrapping a new project. I hope unison will do it more elegantly and with less CPU usage.

EugenMayer:
CPU usage comes from fswatch rather then from the sync itself. I think you will gain a lot of pain with this kind of two way sync with 2 sided fswatch.

Let me state this balsy statement: While i understand every use case is different, i find that code-sharing is always the same. The requirement of 2 way sync is the extra 2% you might want to have, but it comes with a huge minus and is not worth the gain.

As describe here : https://github.com/EugenMayer/docker_sync/issues/20#issuecomment-231582359 you can get it running with rsync very well. We are doing this for a 12k source file project ( 20k in total ) for us, also drupal.

We yet never missed any two-way sync behavior, much more, it would have been an issue with drupal:
 - sites ( user data ) are created in the middle of the codebase ( sites/mysite ) which is a code-base change synced back, while being plain user-data on the container. You simply do not want that on your host ( and work with arbitrary .gitignore entries )
 - You end up syncing back batched assets, uploaded files and all those things, in the middle of your codebase 
 - your codebase can get polluted by a container script without of you wanting that at all - synced back and maybe even committed

With rsync, you use symlinks for sites and store those site-data on a name volume ( or dynamic volume ) to split it from the container.

EugenMayer:
@mickaelperrin i removed your question from the other topic since we will create a huge unreadable thread there.. i did my own mistaked doing that, sorry.

rsync does not use --delete for now, but i worked with that already. I am yet not sure if we should add it to the defaults and make it possible to disable it ( there are some cases were you do not want this, i have such ) or keep the default as such.

I created https://github.com/EugenMayer/docker-sync/issues/37 and happy to hear your input on this

lib/docker_sync/sync_strategy/unison.rb
mickaelperrin: You maybe didn't have remarked that the `chown` is not recursive. The only permission issue that we need to manage manually is the 'base' folder. `unison` handles the permission of all files / directories created inside the 'base' folder with the use of  `user`, `group` and `numericids` options.

lib/docker_sync/sync_strategy/unison.rb
mickaelperrin: I will improve this, as soon as I finish the testing of my unison docker image with `fsevents` built in it.

lib/docker_sync/sync_strategy/unison.rb
EugenMayer: ah, those are very intersting points - i simply was not aware of this, thank you!

EugenMayer:
@mickaelperrin are you aware of https://github.com/EugenMayer/docker-unison ? it would be great if you could build uppon that and have image-based pull requests there, this way also https://github.com/onnimonni/docker-unison can profit from that and we join efforts on the unison approach.

Really tensed to see what you come up with, thank you for your time that you put into this!

mickaelperrin:
Yes, github was down for me when I started working on this and couldn't do a proper fork, but I started my work by using what onnimonni has done.

So far, by making simple files creation / deletion by hand on both sides it works... now it's time for the `composer` test :p

EugenMayer:
@mickaelperrin a pretty interesting POV here https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076/130?u=eugenmayer

I though about it, created a concept: https://github.com/EugenMayer/docker-sync/issues/42

This is probably the better solution, what do you think, in terms of both, performance and robustnes. Share your thoughts on this if you like.

Do not get me wrong, i do not want to defocus you, just avoiding to probably creating work which might again be replaced ( reduction of iterations ). Hope you get this as a contribution

mickaelperrin:
I don't care... what I would love is that someone find a working solution and we could focus on doing things with tools not bulding tools... I would love that your idea will provide a better result than what I am currently experiencing, that would say that we should be near the good solution...

mickaelperrin:
I have opened a new PR https://github.com/EugenMayer/docker-sync/pull/47 that supersedes this one. Maybe we should close this one and continue the discussion in the new one ?

EugenMayer:
Looking at the other PR,  it looks far more superior, i would also tend to close this one. If we want to go back, we can still reopen this one.

PR #22, opened 2016-07-10
[TASK] Unison Remove outdated warn about excludes

mickaelperrin:
Forgot to remove the warning about unsupported excludes in unison


EugenMayer:
ah yes, did overlook this one too, thanks

PR #46, opened 2016-07-13
[FEATURE] Watch only useful events

mickaelperrin:
Even with the `-o` parameter, I found that the watch triggers far more than needed.

This patch limits the events that should be listened by `fsevents` , see http://emcrisostomo.github.io/fswatch/doc/1.9.2/fswatch.html/Invoking-fswatch.html#Filtering-by-Event-Type for a complete list of events.


EugenMayer:
very nice find  ! Did miss that totally, shall be part of 0.0.14

PR #47, opened 2016-07-13
Feature unison fsevents

mickaelperrin:
This PR adds the ability to get files created in the container back to the host immediately without having to wait that a new file is modified on the host side to perform a new sync.

It's intended to work with the PR https://github.com/EugenMayer/docker-unison/pull/3/files


lib/docker-sync/sync_strategy/unison.rb
```diff
@@ -93,11 +120,37 @@ def start_container
...
-        sleep 1
+        sleep 5 # TODO: replace with unison -testserver
         sync
         say_status 'success', 'Unison server started', :green
       end
 
+      def start_local_server
+        say_status 'ok', 'Starting local unison server', :white
+
+        cmd = "unison -socket #{@options['sync_local_server_port']}"
```
EugenMayer: sync_local_server_port seems undocumented?

EugenMayer:
could you merge the current master into your pull request to reduce the changes ( docker-sync )

Could you also add way more inline documentation? Could you explain, which part /component has which function? Thanks

Instead of replacing the current unison-strategy, why not create a new one, unison-XX or something meaningful? This way we could
 - test drive it longer
 - have both strategies provided and keep the normal strategy simple ( also as an example implementation )

mickaelperrin:
Ok I Will do that tomorrow

mickaelperrin:
@EugenMayer, how do you see the use of the docker image ? Should we merge both in one as I did in https://github.com/EugenMayer/docker-unison/pull/3 or create a separate image dedicated to the new strategy ?

EugenMayer:
good question, same thoughts already. Will not have time until today evening to think about that. Maybe a branch, maybe just one image, no idea. What you prefer?

onnimonni:
@mickaelperrin for me the edits which happen inside container will get updated immediately back to my computer after editing them.

How many files do you have in your example project?

mickaelperrin:
As stated in https://github.com/EugenMayer/docker-sync/issues/48, changes / files created / deleted in containers are not synced back to the host until a watch event is triggered on the host.

FYI, I run a project with around 20k files.

EugenMayer:
@mickaelperrin i would go this route:
 - create a new, extra strategy: unison_ng
 - create a long-term-branch on docker-unison and push this a special tag eugenmayer/unison:ng

when someone configures unison_ng as a strategy, we hard pick eugenmayer/unison:ng (no matter what options/unison_image tells us)

One thing is left: Do we need a watch_strategy in this case, or does this needs to be "Dummy" ( no ) - so do we need an OSX watcher?

What do you think about this plan?

EugenMayer:
replaced by https://github.com/EugenMayer/docker-sync/issues/50, closing this one

PR #50, opened 2016-07-14
Strategy unison-dualside with container events

mickaelperrin:
This is a rewrite of the PR #47 that adds the ability to get files created in the container back to the host immediately without having to wait that a new file is modified on the host side to perform a new sync.

It's intended to work with the PR https://github.com/EugenMayer/docker-unison/pull/3/files.

This PR implements : 
- the feature as a new sync strategy
- an example

### DIFF
--- a/example/create-data/Dockerfile
+++ b/example/create-data/Dockerfile
@@ -1,0 +1,6 @@
+FROM alpine
+
+ADD ./entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/example/create-data/entrypoint.sh
+++ b/example/create-data/entrypoint.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+echo "$1" > /command.txt
+
+if [ "$1" == "watch" ]; then
+    adduser -u $(stat -c '%u' /data) -D www-data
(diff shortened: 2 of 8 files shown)


EugenMayer:
exactly what i was thinking about. We might need to add eugenmayer/unison:ng later on, staticly so it matches the more powerful container

EugenMayer:
i like your test case creation, maybe we can do the same for the usual host tests also.
A test script would not thread.join the sync so we can run data-changes and the compare them using either docker exec or docker cp - but thats offtopic here

EugenMayer:
Wonderfull work, awesome! I am sure people will love this new strategy

EugenMayer:
@mickaelperrin add a docker-sync-boilerplate example for unison dualside

EugenMayer:
@mickaelperrin could you double-check https://github.com/EugenMayer/docker-sync/wiki/8.-Strategies since i added a description for the use case for unison-dualside. Be bold to edit / change assumptions and remove any bias-ed opinions i might have added

EugenMayer:
@mickaelperrin i did that, sorry if i am beeing to bold here https://github.com/EugenMayer/docker-sync/commit/2f33d00c9de1a246cd77b7583cbc16bad2fa22de

https://hub.docker.com/r/eugenmayer/unison/tags/

PR #54, opened 2016-07-15
[BUGFIX] strategy needs next version of unison

mickaelperrin:


EugenMayer:
@mickaelperrin shall we rename it form eugenmayer/unison:next to eugenmayer/unison:dualside 

i went boldly and already tagged it on hub.docker.com - if you agree, just change the image name in the code

PR #55, opened 2016-07-15
[BUGFIX] Relative path should be from config file

mickaelperrin:
Relative path sets in the config file seems to be resolved from where the docker-sync process start not from the config file.

This PR forces the relative pathsof `src` to be computed from the path of the `docker-sync.yml` file.

### DIFF
--- a/lib/docker-sync/sync_manager.rb
+++ b/lib/docker-sync/sync_manager.rb
@@ -45,7 +45,7 @@ module Docker_Rsync
       @config_syncs.each do |name, config|
         @config_syncs[name]['config_path'] = @config_path
         # expand the sync source to remove ~ and similar expressions in the path
-        @config_syncs[name]['src'] = File.expand_path(@config_syncs[name]['src'], File.dirname(DockerSyncConfig::project_config_path))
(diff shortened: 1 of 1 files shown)


mickaelperrin:
Mmm, wait it seems there is a problem with that

mickaelperrin:
Seems better now

EugenMayer:
makes a lot of sense when using the config path when resolving relative paths, but as far as i understand, absolute paths would be broken now?

mickaelperrin:
Good question, I didn't check indeed. Personally, I avoid absolute paths for portability issues. But I agree both should work.

mickaelperrin:
I think [expand_path](http://ruby-doc.org/core-1.9.3/File.html#method-c-expand_path) works like [absolute_path](http://ruby-doc.org/core-1.9.3/File.html#method-c-absolute_path), if the main path is absolute it will remain as is.
Did a quick test, and seems to work accordingly.

EugenMayer:
Well when you have a normalized development installation and your run several projects, i can become handy to start / stop something from the other projects, that were "absolute" paths like '~/Development/projects/X' shine.

Thank for testing this!

PR #61, opened 2016-07-16
[FEATURE] Add project-name like -p option of docker-compose

mickaelperrin:
This PR mimics the `-p` option of `docker-compose` by setting a global option `project`.

If set, the `project` value will be append to the container name with an additionnal '_'.

The project name is sanitized by removing all non standard characters.

The special value `auto` determines the project name from the folder name where the docker-sync.yml file is stored.


mickaelperrin:
Doesn't work.

EugenMayer:
I had a similar try with -sync as a suffix, this failed also.

PR #63, opened 2016-07-20
Add terminal-notifier in order to notify when synced

midN:
I don't know if you want to merge it or is it any useful for anybody else.

For me personally i found the need to get notified whenever files were synced so i know i can safely refresh page during development now.
So that's why `terminal-notifier` was added.

It can be made optional if there are some objections though


EugenMayer:
could you explain, what happens, when the notifier is executed? I am absolutely interested in this feature, but could we have it, for now, as an opt int, so creating a sync-option for "notify" and then only do this, when this option is enabled?

EugenMayer:
Ups, forgot something: Thank you for the contribution!

midN:
@EugenMayer - When the notifier is executed, it sends a simple Mac/Apple notification, which looks like this *( Right Top Corner)*:
Notifier

Agreed, sync-option sounds correct way to go, i will add it and update PR

EugenMayer:
eventhough i like the idea, it could actually get a bit noisy, so lets make the option and go with it, great idea having those notifciations

midN:
@EugenMayer - Added it as an config Option, was going to make it as command-line variable, but since they are not Globally defined anywhere it seemed too extreme, so went for option thingy. I added it to `example/docker-sync.yml` as well

VERSION
```diff
@@ -1 +1 @@
-0.0.14
+0.0.15
```
EugenMayer: remove this please

example/docker-sync.yml
```diff
@@ -16,6 +16,9 @@ options:
 syncs:
   # IMPORTANT: this name must be unique and should NOT match your real application container name!
   fullexample-sync:
+    # enable terminal_notifier. On every sync sends a Terminal Notification regarding files being synced. ( Mac Only ).
+    # good thing incase you are developing and want to know exactly when your changes took effect.
+    terminal_notifier: true
```
EugenMayer: please use a semantic name like "notify" here, thanks

example/docker-sync.yml
```diff
@@ -74,4 +77,4 @@ syncs:
     sync_userid: 'from_host'
     unison-dualside_host_ip: 'auto'
     unison-dualside_host_server_port: 6000
-    sync_strategy: 'unison-dualside'
\ No newline at end of file
+    sync_strategy: 'unison-dualside'
```
EugenMayer: could you remove this

EugenMayer:
that looks great already, please just adjust the changes / names and we are good too merge

example/docker-sync.yml
midN: It's a part of `unison-dualside-sync`, didn't change anything here, my vim removed empty-spaces/newline at the end of file, sorry for that :)

midN:
@EugenMayer - Adjusted names and removed `VERSION` change

.gitignore
```diff
@@ -52,3 +52,5 @@ build-iPhoneSimulator/
 # unless supporting rvm < 1.11.0 or doing something fancy, ignore this:
 .rvmrc
 
+# ctags
+tags
```
EugenMayer: why do we need this?

midN: It's great to have `ctags` incase somebody uses them for development, in my case i use it in combination with vim + autocomplete to jump between methods efficiently and autocomplete

EugenMayer: i see, why not

EugenMayer:
great contribution @midN thank you a lot!

Issue #42, opened 2016-07-13
Concept 'magic': Think about implementing a OSXFS strategy architecture for 2 way sync 


EugenMayer:
Mentioned here  https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076/130?u=eugenmayer

Using OSXFS to sync from host to sync container "tmp folder" and then sync from this tmp folder to sync-container app-folder "locally". 

**This has some pros:**
 - fswatch can be removed on host side, saving ressources and "sync delays" (still we need fswatch on the sync container though )
 - fsevents of the host are bubbling to the VM ( no need for custom implementations, OSXFS implements that )
 - it would be probably easier to implement 2 way sync this way

**Cons:**
  a) still needs 2 fswatch processes : One the the sync container to watch the changes on the tmp folder, one on the app-mount folder in the sync container to get app-container changes on the share
  b) docker for mac only and probably OSX only 

**Comments on  the cons:**
a) well, thats just the same number as we would have without OSXFS, so its rather on par, but, fswatch is more effective on linux due to inotify evens, so it might gain performance, sind we reduce the fswatch on OSX, shifting it into linux
b) Well we do support OSX with docker-sync yet only ( i guess, not tested on windows ). Secondly, since it would end up being a strategy only, we could say that some strategies are not available for some docker-types like docker-machine and stuff. So thats sovleable

---

the data flow would look like this
Host --- OSXFS ---> sync-container /tmp[fswatch-trigger]--- local unison sync ---> sync-container /app/code[fswatch-trigger]  ----- named volume mount -> app container

mickaelperrin:
Which tool do you intend to use to to the sync ? `rsync` ?

I tried this approach not with OSXFS but with NFS in the dinghy project, see https://github.com/codekitchen/dinghy/issues/197. The main issue I got was the lack of support of `delete` events in the project `fsevents_to_vm`, using OSXFS mount would indeed help.

EugenMayer:
I would say, this concept is to achieve true transperency for 2 way sync for now - it yet not really change too much for rsync and, since it is docker for mac only, i would not like to switch the current strategies to it, rather add this on as an extra.

I therefor tend to use unison as the local sync. So wie have

Host --- OSXFS ---> sync-container /tmp/ --- local unison sync ---> sync-container /app/code ----- named volume mount -> app container

EugenMayer:
[...]

Issue #51, opened 2016-07-14
Watchers as daemons ?


mickaelperrin:
Hi,

I would love to have the opportunity to start / stop watchers from external calls. This would allow deactivating syncing temporarily when performing some tasks and reenable when needed. 

The first use case of this is currently while installing Drupal from composer, we could wait the end of the bootstrapping script and the run manually a sync task.

What do you think of that ? what about implementing them as [daemons](https://github.com/thuehlinger/daemons) ?

EugenMayer:
Well could you clarify how you run you stack, are you using 

```
docker-sync-stack start
```
 right? Because when using ```docker-sync start``` you can stop/start from an external point by simply ```docker pause <synccontainer-name>``` and then unpause ```docker unpause <synccontainer-name>``` which also would be faster then stopping/starting it.

In general, i like the daemons idea, which also kinda does not block your cli anymore. I am not sure how easy it is since we already run 2 threads, which need to be handled then. Do you have any experience with this kind of implementation in ruby?

EugenMayer:
We could also easy implement this in a separate gem and use docker-sync as a library https://github.com/EugenMayer/docker-sync/wiki/7.-Scripting-with-docker-sync

I do something similar, since after the sync stack starts, i want not only the compose stack to start up, when its started, i want to pre-deploy a developer instance in my drupal. To orchestrate all this, i use a own thor script, docker-sync and docker-compose as libraries

```
thor dk:start --php7
```

This would start a PHP7 driven stack ( compared to the default php5.5 stack ) implemented by overlays with docker-compose-gem, then it pre-installs an instance ( well this is done using consul mainly, but that does not really matter here ).

What i want to say, docker-sync can be used as a lib, also as a lib for a daemon. The question is, if we want the daemon to be an extra-thing without adjusting docker-sync but rather use it as a lib, or if we want to internally change the way docker-sync runs. We might need to both, anyways

midN:
👍  for wrapping whole thing into Daemons and adding option (`-d`) to run it in the background. Was looking into it few days ago, would ideally like to run it in the background and use `terminal-notifier` to let me know it synced. I might look into it on the weekend

[...]

PR #64, opened 2016-07-21
Unison dual-sync feature `unox` for efficient syncing and with user/id mapping

mickaelperrin:
While working on #42, I learned a lot of things about `unison`, some were already pointed partially by contributors (@gagarine and @onnimonni, thanks for making me think about that).

1. There is no absoluty need to run unison with `root` and use `--user` and `--numericids`. Running `unison` under the user we want is sufficient.

2. Unison is very efficient when used with the option `-repeat watch`. It follows events in realtime and update its index in realtime while performing sync. When used in conjunction with `fsevents`, it needs to scan all folder on each request to update it's index.

3. There is no need to run a server on each side. A simple call of unison with `-socket` can handle both sync and watch on **both** sides.

Point 2 and 3 are true, if the system is compatible with `unison-fsmonitor`. On Mac, you need to install [`unox`](https://github.com/hnsl/unox) to get it working.

This PR introduces : 
- a new sync strategy `unison-unox``
- a new watch strategy `unison`
- an example

The adequate `unison` image is [currently here](https://github.com/mickaelperrin/docker-unison-unox). I won't push a PR for `docker-unison` because it is so much simpler than the actual image. Feel free to grab it under your Github account.

I am not sure of the updates of the `execution.rb` and `sync_manager.rb`. I wasn't able to get unison working as a new thread, I needed to fork it. This is a problem because, the `.join` method no longer prevents the main process to exit if you use only this strategy. I use a silly hack with `sleep`, sure there is better implementation than that. But my 1 week experience with ruby prevent me to find a better way ;)

Hope we will get some good results with this new strategy, until the `OSXFS` #42 approach will be patched on the Docker side.


EugenMayer:
Wow, that looks very interesting - thank you for outlining your think-process and way to get this running.

I would love to take some time to have a look at this, to see what you are going to.

Its interesting to see that this is a different go on #42 due to the issues you outlined with OSXFS and events after some time.

Do you already have any benchmarks with unox  vs fswatch on lets say 15k files?

Thank you for your huge effort!

mickaelperrin:
This is the only approach with `magic-sync` that allowed me to scaffold a full Drupal project with 15k in realtime and with reasonable CPU usage.

Unison `dualside` failed with this test and I needed to deactivate syncing to use it while bootstrapping.

I ask me even if this is not faster than `magic` (if we haven't the osxfs issue) as we rely on both side on native file systems.

EugenMayer:
1. So basically unison-dualside fails due to CPU usage on bigger projects ( deal breaker for you and me )

2. And magic fails for now due to the buggy implementation of OSXFS

And this is the third angle to tackle the dual-sync side. Nice progress with those real-case tests

mickaelperrin:
No, dual-side with heavy projects. But you need to deactivate syncing while bootstrapping the project. As soon as the pr

mickaelperrin:
Sorry, I was on the go  while I answered your latest message.

To be precise, `dualside` works with heavy projects, but doesn't support their scaffolding. That's why I implemented `watch_stop` / `watch_start`. It helped me to bootstrap with a script like the one provided on the wiki. 

I would love to see `unox` have a simple run and forget workflow. I will update this PR with my observations while using it.

Would love to have your opinions too and especially on the ruby implementation of the process management.

EugenMayer:
Sure, i will. Will have to probably take on this on the weekend though

mickaelperrin:
Added a new feature that removes the annoying config `sync_host_port`. Now, host port is discovered automatically.

Yet, I don't like parsing the output of cli commands. Maybe, we should think about using a ruby `docker` api, like [`docker-api`](https://github.com/swipely/docker-api) ?

lib/docker-sync/sync_strategy/unison-unox.rb
```diff
@@ -105,7 +106,7 @@ def start_container
           exists = `docker ps --filter "status=exited" --filter "name=#{container_name}" | grep #{container_name}`
           if exists == ''
             say_status 'ok', "creating #{container_name} container", :white if @options['verbose']
-            cmd = "docker run -p '#{@options['sync_host_port']}:#{UNISON_CONTAINER_PORT}' \
+            cmd = "docker run -p '127.0.0.1::#{UNISON_CONTAINER_PORT}' \
```
EugenMayer: looks like 2x :: only one needed

EugenMayer:
i think what you did there is practical, i would not yet pull in a other library ( for now ) if we have more reasons for it, i would follow you on this one. What do you think?

lib/docker-sync/sync_strategy/unison-unox.rb
mickaelperrin: This is intended, I don't specify the host port that should be on the middle of the pattern `ip:host_port:container_port`

EugenMayer: ah, i see, sorry

EugenMayer:
We probably want to thnk about https://github.com/mauricioklein/docker-compose-api since it better parses and utilizes the compose-file and lets us check status / start things by segments. This could make a lot of sense. Did not think deeply into it, pasted for gathering informations

lib/docker-sync/sync_manager.rb
```diff
@@ -121,10 +123,16 @@ def run(sync_name = nil)
...
+        has_threads = false
         @sync_processes.each do |sync_process|
-          sync_process.watch_thread.join unless sync_process.watch_thread.nil?
+          if sync_process.watch_thread
+            sync_process.watch_thread.join
+            has_threads = true
+          end
         end
 
+        sleep unless has_threads
```
EugenMayer: why you want to design this blocking by design?

lib/docker-sync/watch_strategy/unison.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      @options
+      @sync_name
+      @watch_fork
+
+      def initialize(sync_name, options)
+        @options = options
+        @sync_name = sync_name
+        @watch_fork = nil
+        @unison = Docker_Sync::SyncStrategy::Unison_Unox.new(@sync_name, @options)
```
EugenMayer: why do you start the sync strategy once again in the watcher?

EugenMayer:
For now it seems like you pulled way to many changes into this features. You replaced the way threadexec works https://github.com/mickaelperrin/docker-sync/blob/feature/unox/lib/docker-sync/execution.rb#L24 which needs testing for itself. I would suggest to make this a own PR we can test isolated since this will probably influence all fswatch or at least, ensure fswatch still works like desired.

In general, lets try to now blow this out of proportion or we will have heard time to get it all together. It seems like you try to already implement the "daemon" mode here , mentioned in #51  

To be reviewed / documented:

- new join_threads + sleep in there
- new forkexec
- WatchStrategy::Unison::initialize -> why is a sync strategy started here? Is this the "host watcher?"

In general could you document what you did in SyncStrategy::Unison_Unox so what is started what for, why which arguments - so we have some knowledge sharing and can maintain this easier in the future ( or other contributors ). Thanks!

lib/docker-sync/sync_manager.rb
mickaelperrin: As I said previously in the first message of the PR.

> I wasn't able to get unison working as a new thread, I needed to fork it. This is a problem because, the .join method no longer prevents the main process to exit if you use only this strategy. I use a silly hack with sleep, sure there is better implementation than that.

If you don't sleep, the main process exits and you have no way to kill started unison process automatically.

lib/docker-sync/watch_strategy/unison.rb
mickaelperrin: This doesn't start the `sync` process, this only instantiate the same class.

Basically running the watcher is the same command as the sync but with the `-repeat watch` option. To ensure that all options are well used, reuse a maximum of code and stick to the strategy pattern, I thought it was the best way to do it.

mickaelperrin:
Well spotted for the `threadexec`, this is a bad commit. This function doesn't have to change in any way. I will update this.

lib/docker-sync/watch_strategy/unison.rb
EugenMayer: Makesa lot of sense - just add an inline comment and this is great, thanks

lib/docker-sync/sync_manager.rb
EugenMayer: well we have to start thinking if we start working with pid / pid files whatever solutions are common in ruby.

mickaelperrin:
Need your help on this.

This morning, this stopped working on my machine. No sync happens  even if the sync folder has only very limited files.

The issue is really weird... it's related to the command `unison ... -repeat watch`. 

When this command is launched through `docker-sync` nothing happens, but if I start the container manually and copy / paste the command, it works...

This is the same issue that makes me switch from `Thread.new` to `Process.fork` but the trick seems to stop working...

Do you have some any ideas why a process started through `ruby` coud be different than one launched manually from the command line ?

EugenMayer:
This most probably has to do with the "cli_mode" detection, this means, either it shall run thor sync:sync or docker-sync sync

Especially if you run in scaffolding and use thor, things go mad. If you run scaffolding and you utilize docker-sync as lib and also run it of course, as a cli tool with the watcher, you need to force the cli_mode to "docker-sync" https://github.com/EugenMayer/docker-sync/blob/master/example/docker-sync.yml#L17

If this sounds confusing, please see the code here https://github.com/EugenMayer/docker-sync/blob/master/lib/docker-sync/watch_strategy/fswatch.rb#L62 it might explain it better.

Most probably the watcher calls thor sync:sync not docker-sync sync in your case, but thor sync:sync does not exists in your thor implementation of course

mickaelperrin:
Thanks for your quick answer. However, it looks like it is not related to what you spotted.

Current tests are run by manually launching `docker-sync` from `cli`. After each change, I run the `deploy_locally.sh` script to rebuild the gem. I forced the `cli_mode` parameter to `docker-sync` as you spotted but it didn't change anyhting.

in fact, `sync` works well. When I launch the `docker-sync start` command, the initial sync is performing well, the watch is launched also I can see it in the process. But it doesn't respond to file events... If I kill the process, start the container manually, copy and paste the watch command echoed, then it works...

EugenMayer:
Is this unox related or fswatch related? If the latter, lets pickup a new issue. we are using fswatch in production right now and at least several people do no have issues here.

if it is unox related, and "fileevents" you are actually the expert for now - or did i miss the point?

mickaelperrin:
:) This is only `unox` related indeed.

But, my question is indeed more related to general ruby *why a process launched in ruby doesn't perform similarly as a process launched from cli* ? should I take attention of some thing in particular.

onnimonni:
Could it be that the ruby process doesn't have unox in it's `$PATH`?

EugenMayer:
lets split of the general question then into a special issue, what do you think? The reason for this issue could be on different layers, we have to envestigate. ENV like path could be a cause

lib/docker-sync/sync_strategy/unison-unox.rb
```diff
@@ -0,0 +1,204 @@
...
+        @sync_name = sync_name
+        # if a custom image is set, apply it
+        if @options.key?('image')
+          @docker_image = @options['image']
+        else
+          @docker_image = 'eugenmayer/unison:unox'
+        end
+        begin
+          Preconditions::unison_available
+          Preconditions::unox_available
```
mickaelperrin: @onnimonni thanks for trying to help, but I have implemented a precondition to check that `unox` is accessible from the `docker-sync` process.

mickaelperrin: @onnimonni thanks for trying to help, but I have implemented a precondition to check that `unox` is available. Or maybe this check is broken ?

mickaelperrin:
Using backticks instead of `popen3`did the trick. Now, need to understand why...

EugenMayer:
You probably want to have a look at http://i.stack.imgur.com/1Vuvp.png .. 

So the main differences between popen3 and backticks are, popen is non blocking, backticks are blocking. 

The change you just made will not make it into the master branch, since this once again influences fswatch behavior. Please either create a customer execution or solve the issue a different way. We wont stderr/out to be captured - so i suggest you revert that

mickaelperrin:
I agree, I wish it has a "better" solution. But to be precise:

- this won't influence `fswatch` as this is implemented in the function `forkexec` that it called with only this watch strategy
- stdout is captured, as indicated in your graphic. If we need stderr, I think adding `2&>1` to the command will help.

EugenMayer:
oh sorry, got confused, my bad. I guess we will have to solve a couple of those  cli-invocation issues somehow, e.g. a issue with process #69 or #65 . Probably a total different approach as you suggest either kind of a process manager or somewhat different. The issue will be, not to "overcomplicate" things.

I do not have a slighly idea why backticks work, popen do not. I those cases, it was always the ENV being different or one of the commands is pickier about what is needed to be escaped?

mickaelperrin:
I checked that using backticks still ensure that processes are killed when we kill the main process. But maybe, in some circumstances, we will have some zombie processes as you get with `fswatch` in #69 .

Regarding implementation as a daemon #69 , I think it would be a good enhancement. But to be honest, I need less this feature since the use of `unox`. As it performs well so far, I let it run even when doing heavy I/O. If something goes wrong, kill and restart correct synchronisation issues.

However from my point of view both #69 and #65 are independent of this PR at the moment. But, yes this PR will need to be updated if we change someting more globally.

EugenMayer:
agreed, independent but it could solve issues occurring here probably, thats why i referenced here, so we could probably keep track of the progress. I just expect that we go different routes solving similar issues and see, what solutions works out best

mickaelperrin:
@EugenMayer I just let you review the latest commits and will merge the branch if it is ok for you.

lib/docker-sync/preconditions.rb
```diff
@@ -43,4 +43,12 @@ def self.unison_available
       raise('Could not find unison binary in path. Please install it, e.g. using "brew install unison"')
     end
   end
+
+  def self.unox_available
+    if (find_executable0 'unison-fsmonitor').nil?
+      raise('Could not find unison-fsmonitor binary in path. Please install it, see https://github.com/hnsl/unox, or simply run :
+      wget -O /usr/local/bin/unison-fsmonitor https://raw.githubusercontent.com/hnsl/unox/master/unox.py \
+      && chmod +x /usr/local/bin/unison-fsmonitor')
+    end
```
EugenMayer: @mickaelperrin a pitty there is now brew alternatives. I am thinking about creating a own tap and adding this.

Actually, in this case, use yes? and ask if the command should be run, run it automatically then. This makes the most sense here to ease up the installation

EugenMayer:
@mickaelperrin for my perspective, this looks great. The one thing automating / easing up unox installation with a confirmation would actually help a lot and then just head on mergin it. Great work

EugenMayer:
@mickaelperrin see #80 , we should change back the image name to eugenmayer/unison:unox due to legacy support / pre release support . We chnage it to eugenmayer/unison some releases later, so we are sure no/few are using the older versions.

lib/docker-sync/preconditions.rb
mickaelperrin: I thought about auto install also, but as you didn't provide autoinstall of unison / fswatch / rsync... I stick with a simple help message on how to install.

EugenMayer: The reason was, not everybody has brew. In your case, no dependencys are needed and its the only way to install. AFAIK wget is not part of OSX by default, you might want to use curl.

I will change the others also by the way, adding a question, like for the images

mickaelperrin: Done

lib/docker-sync/preconditions.rb
```diff
@@ -46,9 +46,14 @@ def self.unison_available
...
-      curl "https://raw.githubusercontent.com/hnsl/unox/master/unox.py" -o "/usr/local/bin/unison-fsmonitor" \
-      && chmod +x /usr/local/bin/unison-fsmonitor')
+      cmd = 'curl "https://raw.githubusercontent.com/hnsl/unox/master/unox.py" -o "/usr/local/bin/unison-fsmonitor" \
+      && chmod +x /usr/local/bin/unison-fsmonitor'
+      Thor::Shell::Basic.new.say_status 'warning', "Could not find unison-fsmonitor binary in path. Please install unox before you continue, see https://github.com/hnsl/unox.", :yellow
+      if Thor::Shell::Basic.new.yes?("Shall I install unison-fsmonitor for you?")
+        `#{cmd}`
+      else
+        raise("Please install it, see https://github.com/hnsl/unox, or simply run :\n #{cmd}")
+      end
```
EugenMayer: perfect

lib/docker-sync/sync_strategy/unison.rb
```diff
@@ -23,7 +23,7 @@ def initialize(sync_name, options)
         if @options.key?('image')
           @docker_image = @options['image']
         else
-          @docker_image = 'eugenmayer/unison'
+          @docker_image = 'eugenmayer/unison:unox'
```
EugenMayer: thanks, that makes our live a bit easier regarding the upgrade process

Issue #87, opened 2016-08-02
Sync containers may fail to start depending on their name


mickaelperrin:
Say you have a running sync container called `project-data`, if you try to create a sync container called `project` or `data` it will fail saying that the container is already running.

EugenMayer:
I think something is wrong in the description. You mean, you cannot name an app container like a sync container, right? Yes, this is correct and we cannot absoltely change that. If you start suffixing the container name, you will fail on docker-sync sync, since you cannot recall your old container name anymore. We would need to adjust here a lot more. Maybe all this becomes possible when we have a daemon mode or similar

EugenMayer:
(please do not merge this for now)

EugenMayer:
sorry, was confused, that makes total sense ( its about partial name match). I merged it, closing this issue. Thanks

PR #88, opened 2016-08-02
[BUGFIX #87] Container may fails to start

mickaelperrin:
- PB: depending on their name some container may fails to start. This
comes from the way to test if a container exists / is already running.
- FIX: extract the container name from the docker inspect command
and sed and ensure it matches with grep on full string


EugenMayer:
oh i totally missunderstood this issue, i am very sorry. You are talking about a partial match of course. That makes total sense!

Issue #84, opened 2016-08-02
How "-prefer newer" works


gagarine:
The `-prefer newer` unison's flag is based on files date? The code-sync containers doesn't have the same time zone than my local machine. Can this cause a problem?

EugenMayer:
@mickaelperrin you are the expert here, cannot judge uppon this. 

I would suggest it matters of course.

gagarine:
This can explain why I was getting conflict.

Perhaps we should start the docker-sync container with
`-v /etc/localtime:/etc/localtime:ro`

EugenMayer:
Interesting point, you mean, setting the sync-container time to the one of OSX. Maybe instead of mounting it, we can pass a ENV var, an read it before we start the container or even make this a configuration variable ( in addition )

mickaelperrin:
I didn't find any documentation about how `-prefer newer` works.  I will give a look to the C code but actually we should ensure that the unison container share the same time and timezone with the host.

I will provide a patch to ensure this.

PR #89, opened 2016-08-02
[FEATURE #84] Pass the timezone of host as env var

mickaelperrin:
Grabs the timezone of the host and pass it to the container in an environment variable.

Note: this is not sufficient and needs an adjustment in the docker entrypoint script.

see PR in related projects


EugenMayer:
we have to rebuild and push the images to have this working, see #90

PR #95, opened 2016-08-09
[BUGFIX] osx vanilla sed doesn't support the -r option

mickaelperrin:
- PB: on some systems sed doesn't support the exended syntax.
- FIX: rewrite the regexp without the use of the extended syntax


EugenMayer:
great stuff, so we do not need to let the user upgrade sed

Issue #103, opened 2016-08-09
Unison invoked incorrectly too many roots


timstallmann:
Just started seeing this issue on the 0.1.0 branch. I updated my docker sync config per the migration docs.

`docker-sync.yml` contents:
```
syncs:
  site_www-sync:
    src: './'
    sync_host_ip: 'localhost'
    dest: '/var/www/html'
    sync_strategy: 'unison'
```

`docker ps output`:
```
f21d0fac8d2c        eugenmayer/unison:unox   "/entrypoint.sh super"   25 seconds ago      Up 25 seconds       127.0.0.1:32770->5000/tcp   site_www-sync
```

I've tried this with and without the `sync_host_ip` and that flag doesn't seem to impact the performance

mickaelperrin:
Could you paste the output of `docker-sync start` ?

FYI, if not provided sync_host_ip is by default 'localhost'.

timstallmann:
Sure! This is with verbose: true:

```
          ok  Starting unison
          ok  site_www-sync container not running
          ok  creating site_www-sync container
     command  docker run -p '127.0.0.1::5000'                               -v site_www-sync:/var/www/html                               -e UNISON_DIR=/var/www/html                               -e TZ=${TZ-`readlink /etc/localtime | sed -e 's,/usr/share/zoneinfo/,,'`}                                                                                             --name site_www-sync                               -d eugenmayer/unison:unox
          ok  starting initial sync of site_www-sync
     command  docker inspect --format=" {{ .NetworkSettings.Ports }} " duketoday_www-sync | sed "s/.*map\[5000[^ ]\+\s\([0-9]\+\).*/\1/"
     command  unison /Users/tim/htdocs/site -auto -batch socket://localhost: map[5000/tcp:[{127.0.0.1 32770}]]
       error  Error starting sync, exit code 0
     message  Usage: unison [options]
    or unison root1 root2 [options]
    or unison profilename [options]

For a list of options, type "unison -help".
For a tutorial on basic usage, type "unison -doc tutorial".
For other documentation, type "unison -doc topics".
```

mickaelperrin:
Thanks got it, there is a problem with `sed`. Could you paste the result of `sed --version` ?

timstallmann:
sw_vers -ProductVersion is 10.11.6.. `sed --version` gives an illegal option error, as does `sed -V`

mickaelperrin:
[...]

PR #104, opened 2016-08-09
[BUGFIX #103] Unison reports too many roots

mickaelperrin:
- PB: the sed expression to grab the used port in container is not
compatible with the default sed binary.
- FIX: force the use of the default sed binary /usr/bin/sed and update
the regexp accordingly


EugenMayer:
@mickaelperrin good job, also for reacting that quick. Releasing 0.1.1 right away

PR #113, opened 2016-08-17
Specify gemspec in Gemfile

fidothe:
Bundler can read dependencies directly from the .gemspec, which reduces
duplication and potential dependency masking problems

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,3 @@
 source 'https://rubygems.org'
-gem 'thor'
-# compose is a dynamic dependency, if someone is not using stack, he will not need it.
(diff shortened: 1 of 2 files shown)


EugenMayer:
Wow, that is something new for me! How about IDE integrations trying to parse Gemfile and load the Gems missing, lets say IntelliJ?

fidothe:
It's the canonical way to manage gem dependencies in the development of a gem. Certainly RubyMine just uses Bundler here, so it's not a problem

EugenMayer:
Thank you for clarifying, nice way to learning something new :)

PR #121, opened 2016-08-20
[BUGFIX] Use sync_host_ip for docker run command

BKStephens:
Was hardcoded to 127.0.0.1. This will work great if sync_host_ip is localhost,
but will not work if you are using docker-machine and sync_host_ip is something else

### DIFF
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -134,7 +134,7 @@ module Docker_Sync
           if exists == ''
             say_status 'ok', "creating #{container_name} container", :white if @options['verbose']
             run_privileged = '--privileged' if @options.key?('max_inotify_watches') #TODO: replace by the minimum capabilities required
-            cmd = "docker run -p '127.0.0.1::#{UNISON_CONTAINER_PORT}' \
+            cmd = "docker run -p '#{@options['sync_host_ip']}::#{UNISON_CONTAINER_PORT}' \
                               -v #{volume_name}:#{@options['dest']} \
                               -e UNISON_DIR=#{@options['dest']} \
                               -e TZ=${TZ-`readlink /etc/localtime | sed -e 's,/usr/share/zoneinfo/,,'`} \


mickaelperrin:
Thx for the PR, I think you're right. I didn't take in consideration that some users are not using the native Docker for Mac.

FMI, which solution are you using ? Why don't you use Docker for Mac ?

EugenMayer:
This makes absolute sense. 

But to not break the old behavior, please allow the #{@options['sync_host_ip']} to be not set and default it to 127.0.0.1 if it is unset. This way we support the old "less configuration way" but still can support any other docker machine

mickaelperrin:
@EugenMayer But is it not already the case with https://github.com/EugenMayer/docker-sync/blob/c4e9f21bc4c6510e8186e25765a1f786e31ad470/lib/docker-sync/sync_process.rb#L24

EugenMayer:
@mickaelperrin hmm you might be right on this. Should we change the default from localhost to 127.0.0.1 - does this matter for unison ( like it matters for e.g. mysql )  .. socket connectionss

mickaelperrin:
Indeed, I remember having read some troubles about using 127.0.0.1 instead of localhost or vice-versa. But, to be honest, I don't know if it could impact anything here.

However, some projects on Girhub are using `localhost` instead of `127.0.0.1` for port mapping. So, maybe, we should stick with with the current default localhost` used everywhere else where we call `#{@options['sync_host_ip']}`.

EugenMayer:
ok, lets stay with localhost and see what happens, you re right.

BKStephens:
@mickaelperrin In regards to https://github.com/EugenMayer/docker-sync/pull/121#issuecomment-241186938, I am currently using docker-machine, but have tried out Docker for Mac. However, I think I am running into this bug with Docker for Mac: https://github.com/docker/compose/issues/3277

Sometimes Docker for Mac will start and everything will work, but half of the time it tells me a port is already allocated. Sometimes 6379 (redis), sometimes 5432 (postgres).

When this happens, I see what is listening with:
```
 sudo lsof -i TCP:5432 | grep LISTEN
```
Then kill the process that is listening. I can run the command again to make sure it is gone.

Then, I try `docker-compose up` again and I see the same error. Occasionally quitting Docker for Mac and starting it again will resolve the issue, but most of the time I have to restart my Mac. I tried uninstalling Docker for Mac and reinstalling it, but that did not help. This bug is kinda a deal breaker, so I plan on staying on docker-machine for a little while longer.

EugenMayer:
@BKStephens eventhough that is not a docker-compose issue queue, you issue is 99% sure not related to docker-for-mac, but either compose itself or the way you stop services in general.

We do run a gazzilian services, up and down, stop , start whatever, on the same ports, and it always works.

Nevertheless, docker-sync shall not force you in any of the docker-machine implementations, so this issue will be solved with the next release

ZZromanZZ:
@EugenMayer Do you plan stable release with this fix ? It is quite a pain to work with development branch currently.

EugenMayer:
@ZZromanZZ no plans yet, but i try to get it out this week, since i am on hollidays for to weeks, starting the next week

EugenMayer:
switched to 127.0.0.1 with 0.1.2 since i identified corner cases where localhost fails, see #134

PR #129, opened 2016-08-29
fix container detection

rogamoore:
This should fix EugenMayer#111 and correctly recognize running or existing containers and proceed accordingly instead of always trying to create a new container.


EugenMayer:
very well done @rogamoore, thank you a lot!

PR #139, opened 2016-09-07
Support configuring rsync allowed hosts and copying symlinks

yacn:
These features were needed for us to get docker-sync working in our environment.

Added the option `rsync_allowed_cidr` where you can specify a string of CIDRs rsync allows to sync with, e.g. ours looks like:

```yaml
....
syncs:
  foo-sync:
    ...
    rsync_allowed_cidr: '10.10.10.0/24 192.168.0.0/16 172.16.0.0/12'
    copy_symlinks: 'true'
...
```

This PR also adds the option `copy_symlinks` which adds the `-L` option when syncing code into the container. We were running into a permissions issue when trying to copy the symlink directly into the container.

### DIFF
--- a/docker-sync.gemspec
+++ b/docker-sync.gemspec
@@ -12,6 +12,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.0'
   s.add_runtime_dependency 'thor', '~> 0.19', '>= 0.19.0'
   s.add_runtime_dependency 'gem_update_checker', '~> 0.2.0', '>= 0.2.0'
-  s.add_runtime_dependency 'docker-compose', '0.8.3'
+  s.add_runtime_dependency 'docker-compose', '0.8.4'
   s.add_runtime_dependency 'terminal-notifier', '1.6.3'
 end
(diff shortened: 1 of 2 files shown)


yacn:
Ping @EugenMayer

EugenMayer:
Thank you for the contribution!Cidr is not needed, since we expose the port, no limits applied? Copy symlinks needs to be reviewed, yet we had adjustments on the server side.

yacn:
I'll give it a try without the CIDR configured, you may be right that it's not needed and the only thing we needed was `copy_symlinks`.

EugenMayer:
please also remove the docker-compose version bump - generally do not stuff 3 features into one pull request, that makes things a lot easier to handle :)

Upgrading docker-compose will be part of the next release anyway, but we will jump on 1.0.0 acually

EugenMayer:
@yacn news on this? you will need to rebase to master, since the gemspec file changed

yacn:
@EugenMayer Sorry, we moved over to using `docker-machine` at work so this has kind of fallen behind. I'll clean up, remove the unneeded stuff and rebase tomorrow morning.

EugenMayer:
@yacn this means you do no longer need docker-sync or rather this change?

PR #141, opened 2016-09-15
configuration to pass --build flag to `docker-compose up`

dylanscott:
Thanks for all your work on docker-sync! It's been really useful in speeding up my local development process.

One thing I was missing from my purely `docker-compose` based process was being able to pass the `--build` flag to `docker-compose up` to ensure that any files not synced from my host into the container were up-to-date.

I just submitted [a change](https://github.com/xeger/docker-compose/pull/39) to the docker-compose gem to expose the `--build` flag, and I was hoping you'd consider adding a configuration for docker-sync to pass this flag. Currently that flag is called `compose-force-build` in the global options section, but I'd be happy to change it if you have any suggestions.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     docker-sync (0.1.2)
-      docker-compose (= 0.8.3)
+      docker-compose (= 1.0.0)
       gem_update_checker (~> 0.2.0, >= 0.2.0)
       terminal-notifier (= 1.6.3)
       thor (~> 0.19, >= 0.19.0)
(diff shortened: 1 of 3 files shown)


EugenMayer:
please remove all changes to the gem-dependencies and also make the parameter optional, so check if it is set and only pass it, if it does. Thank you!

dylanscott:
I just pushed a change to only pass the `build` parameter to `@compose_session.up` if the parameter is set. I haven't written much ruby though so let me know if there's a more idiomatic way to do it.

Regarding the gem-dependencies, I probably should have been more explicit. The change to support the `build` flag on the `docker-compose` side is only included in the [1.0.0](https://github.com/xeger/docker-compose/commits/v1.0.0) release, so I think this change won't work without the dependency bump.

EugenMayer:
great stuff, thank you a lot, also for checking with the docker-compose developer!

dylanscott:
Awesome, thank you!

mrfelton:
This could do with being documented

PR #165, opened 2016-10-31
Fix typo in command output

GordonDiggs:
Noticed this minor typo while I was using docker-sync


EugenMayer:
good catch, thank you!

PR #166, opened 2016-11-01
Don't hardcode the path for sed - e.g. sed is in /bin on most Linux systems

marmarama:
I have to support a Docker dev environment where the developers use a mixture of macOS and Linux.  I use docker-sync with the unison sync strategy, because the performance of the dev environment using Docker for mac osxfs shares is terrible.  I also don't want to have to maintain two completely separate paths for macOS and Linux workstations, so I install and use docker-sync on both platforms.

It actually works very well on Linux, with very little fiddling required, as long as the unison deb/RPM includes the Linux native 'unison-fsmonitor'.  However, there is one little gotcha - the unison sync strategy hardcodes the path to sed as /usr/bin/sed, which isn't where it is located on Linux - it's in /bin/sed usually.

So, don't hardcode the path - let $PATH do its job.

### DIFF
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -173,7 +173,8 @@ module Docker_Sync
       end
 
       def get_host_port(container_name, container_port)
-        cmd = 'docker inspect --format=" {{ .NetworkSettings.Ports }} " ' + container_name + ' | sed  -E "s/.*map\[' + container_port + '[^ ]+ ([0-9]*)[^0-9].*/\1/"'
+        File.exist?('/usr/bin/sed') ? sed = '/usr/bin/sed' : sed = `which sed`.chomp # use macOS native sed in /usr/bin/sed first, fallback to sed in $PATH if it's not there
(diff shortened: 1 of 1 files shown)


mickaelperrin:
We used to use the `sed` path defined in $PATH. But, on OSX, multiple sed can be installed (native and GNU version) that doesn't use the same arguments. As far as I remember, we force the path to ensure the native version is used.

So maybe, we should introduce a parameter that define the path to sed ? Or check the environment to define it dynamically ?

marmarama:
Either approach works for me.  However, if the issue is with GNU sed, I can say that the sed command works fine in GNU sed 4.2.2: that's what's installed on my Ubuntu test box.  GNU sed supports '-E' as an undocumented option for compatibility with BSD sed, so it treats it as an extended regex correctly.

I guess an alternative approach would be to use a Ruby regex instead of sed.

EugenMayer:
Sure, i am fine with defining an absolute, configureable path to sed - the reason to hardcode is like @mickaelperrin mentioned.

Are you keen enough to provide a pull reqest @marmarama ? We love contributions :)

marmarama:
Updated.  Now it tries to find /usr/bin/sed and uses that first, but if /usr/bin/sed doesn't exist, it uses the sed in $PATH.  Same behaviour as before on macOS, but working fallback on other platforms.

EugenMayer:
Makes a lot of sense, @mickaelperrin any vetos?

EugenMayer:
Thank you @marmarama , thats a great addition!

Issue #130, opened 2016-08-29
Only check for update for the current strategy


rogamoore:
Currently when docker-sync checks for an updated image it will check all images instead of just the one that's needed. It will download the first new image, then stop and ask the user to run `docker-sync clean`. After that it will download the second new image, then stop and ask the user to run `docker-sync clean` and so on. In total you have to run `docker-sync start` followed by `docker-sync clean` about three times.

```
roga:~/code/]$ docker-sync start
          ok  Checking if a newer rsync image is available
Status: Downloaded newer image for eugenmayer/rsync:latest
     warning  Downloaded newer image for rsync
     warning  Please use "docker-sync clean" before you start docker-sync again
[roga:~/code/]$ docker-sync clean
Error response from daemon: No such container: appcode-unison-sync
Error response from daemon: No such container: appcode-unison-sync
Error response from daemon: get appcode-unison-sync: no such volume
     success  Finished cleanup. Removed stopped, removed sync container and removed there volumes
[roga:~/code/]$ docker-sync start
          ok  Checking if a newer rsync image is available
     success  Image is (now) up to date
          ok  Checking if a newer unison:onesided image is available
Status: Downloaded newer image for eugenmayer/unison:onesided
     warning  Downloaded newer image for unison:onesided
     warning  Please use "docker-sync clean" before you start docker-sync again
[roga:~/code/]$
```

Any idea how to do all this in one step or just pull the images that are really needed? I'm never using rsync for example so I don't see why it should be downloaded.

EugenMayer:
Yes that would be a good improvement, admitted. Feel free to create a pull request if you are able to work on this, can take me a while to implement this, since this is at least "cosmetic" for now

rogamoore:
I just opened a PR :)

EugenMayer:
Thank you a lot @rogamoore very nice contribution and find!

PR #168, opened 2016-11-07
simplify check for new images

rogamoore:
This PR will fix https://github.com/EugenMayer/docker-sync/issues/130 by checking if there are updated docker images available in one single step instead of stopping after each check.


EugenMayer:
that makes a lot of sense! thank you

PR #173, opened 2016-11-14
remove rsync trailing slash in command args

dojohnso:
Since the docker-sync.yml mentions the user has the option to add a
slash or not, we should not hardcode a trailing slash.

also fixed a typo in the yml (tilde)

### DIFF
--- a/example/docker-sync.yml
+++ b/example/docker-sync.yml
@@ -21,7 +21,7 @@ syncs:
     # enable terminal_notifier. On every sync sends a Terminal Notification regarding files being synced. ( Mac Only ).
     # good thing incase you are developing and want to know exactly when your changes took effect.
     notify_terminal: true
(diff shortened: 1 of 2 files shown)


EugenMayer:
@dojohnso that sounds all very reasonable. The only bad feeling i have now is, that actually this wil change the sync behavior for all yet not used the slash. Releasing that in the patch version make me think.

dojohnso:
@EugenMayer that's a good point. what do you think about doing this instead: check the 'src' for a period (.) and if found, assume it's a filename.ext (so a single file) and don't add a trailing slash. But if not found, add the slash?

Although that isn't foolproof since directories could have periods in them. Is there a way you can make sure people that update know to double check their src value? Based on the comments in the yml I'd expect them to be using it or not as desired since it implies it takes it into consideration.

EugenMayer:
Well i think you are right, i will add a migration node but actually it always have been there, that there is a difference. Thank you for wrapping your head around this and of course for your contribution!

PR #179, opened 2016-11-22
Rsync single file

dojohnso:
Added some example settings for src/dest settings for full directory vs
a single file

### DIFF
--- a/example/docker-sync.yml
+++ b/example/docker-sync.yml
@@ -21,8 +21,16 @@ syncs:
     # enable terminal_notifier. On every sync sends a Terminal Notification regarding files being synced. ( Mac Only ).
     # good thing incase you are developing and want to know exactly when your changes took effect.
     notify_terminal: true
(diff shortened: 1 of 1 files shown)


EugenMayer:
i the * mandatory to sync a folder ?

dojohnso:
@EugenMayer yeah for whatever reason, when I don't use the "path/*" format, it creates a "path/path/" directory on the application image. i haven't dug into why too much. i just know that /* works. fswatch seems works fine with that as well.

EugenMayer:
Well i recall that was the reason we auto added the slash.. hmm that will become an ugly upgrade bug

dojohnso:
@EugenMayer when i just use "/this/path/", the fswatch output says:

> Synced /this/path
> Starting to watch /this/path- Press CTRL-C to stop

note it doesn't show the slash there. so it looks like something about the code needs the * to work as intended

dojohnso:
@EugenMayer although I'm playing with some different configurations that might not require the *. I'll keep you posted.

EugenMayer:
Thank you @dojohnso !

PR #188, opened 2016-11-30
Preserve forward slash in sync src after expanding path

nbr:
#173 causes a regression which affects sync behavior in at least the rsync strategy.

The use of `/` in rsync is signficant, and results in different behaviors. Example:
```
$ tree
├── a
│   ├── 1
│   └── 2
├── b
├── c
$ rsync -a a b; tree
├── a
│   ├── 1
│   └── 2
├── b
│   └── a
│       ├── 1
│       └── 2
├── c
$ rsync -a a/ c; tree
├── a
│   ├── 1
│   └── 2
├── b
│   └── a
│       ├── 1
│       └── 2
├── c
│   ├── 1
│   └── 2
```

`File#expand_path` is called on the `src` to support relative path constructs. Previously, a forward slash (`/`) was hardcoded into the result. #173 removed this hardcoded `/`. `File#expand_path` removes `/`, thus making it impossible to pass `src` containing a `/` to eventually be given to sync `cmd` as an argument. This prevent us from being able to sync directories as was done between `a` and `c`

### DIFF
--- a/lib/docker-sync/sync_manager.rb
+++ b/lib/docker-sync/sync_manager.rb
@@ -44,9 +44,14 @@ module Docker_sync
     def upgrade_syncs_config
       @config_syncs.each do |name, config|
         @config_syncs[name]['config_path'] = @config_path
-        # expand the sync source to remove ~ and similar expressions in the path
-        @config_syncs[name]['src'] = File.expand_path(@config_syncs[name]['src'], File.dirname(@config_path))
-        @config_syncs[name]['cli_mode'] = @config_global['cli_mode'] || 'auto'
+
+	# [nbr] convert the sync src from relative to absolute path
+	#	preserve '/' as it may be significant to the sync cmd
+        absolute_path = File.expand_path(@config_syncs[name]['src'])
+	absolute_path << "/" if @config_syncs[name]['src'].end_with?("/")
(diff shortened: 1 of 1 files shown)


lib/docker-sync/sync_manager.rb
```diff
@@ -44,9 +44,14 @@ def get_sync_points
     def upgrade_syncs_config
       @config_syncs.each do |name, config|
         @config_syncs[name]['config_path'] = @config_path
-        # expand the sync source to remove ~ and similar expressions in the path
-        @config_syncs[name]['src'] = File.expand_path(@config_syncs[name]['src'], File.dirname(@config_path))
-        @config_syncs[name]['cli_mode'] = @config_global['cli_mode'] || 'auto'
+
+	# [nbr] convert the sync src from relative to absolute path
+	#	preserve '/' as it may be significant to the sync cmd
+        absolute_path = File.expand_path(@config_syncs[name]['src'])
```
EugenMayer: please fix the indent, thanks!

EugenMayer:
That is a really good find and it was not intended to remove / forced, rather the other way around, / should not be added to the path by force - that is what 173 tried to fix. Thank you a lot for this brief explanation and also the fix, very appreciated!

Valuable contribution!

PR #197, opened 2016-12-12
Change docker-compose req to '~> 1.0.2' in gemspec

fdintino:
There is a [1.4.0.alpha1 version of docker-compose on rubygems](https://rubygems.org/gems/docker-compose/versions/1.4.0.alpha1) that is completely incompatible with the 1.0.x API. Oddly, it is more than a year older than the most recent 1.0.x release (1.0.3), so I'm assuming that it is some sort of mistake. I've changed the gemspec version constraint to prevent that release from being installed.


EugenMayer:
I will talk to the DC author, lets see what it is. In any way, thank you for pointing at this, pinning will make sense!

EugenMayer:
Not needed anymore, the DC author removed the released gem, it was a accident

PR #222, opened 2017-01-25
Support arrays in compose file paths

codycraven:
Allow docker-sync.yml to define an array or a string when defining files for `docker-compose up` in `docker-sync-stack` (without breaking existing functionality).

Examples:
```yml
options:
  compose-dev-file-path:
    - 'docker-compose.local.yml'
    - 'docker-compose.local.mac.yml'
```

```yml
options:
  compose-file-path:
    - 'docker-compose.yml'
    - 'docker-compose.local.yml'
    - 'docker-compose.local.mac.yml'
```

```yml
options:
  compose-file-path:
    - 'docker-compose.yml'
    - 'docker-compose.local.yml'
  compose-dev-file-path: 'docker-compose.local.mac.yml'
```

Supports [docker-compose's multiple -f](https://docs.docker.com/compose/reference/overview/) options.

### DIFF
--- a/lib/docker-sync/compose.rb
+++ b/lib/docker-sync/compose.rb
@@ -10,22 +10,15 @@ class ComposeManager
     ### production docker-compose.yml
     compose_files = [File.expand_path('docker-compose.yml')]
     if @global_options.key?('compose-file-path')
-      path = File.expand_path(@global_options['compose-file-path'])
-      unless File.exist?(path)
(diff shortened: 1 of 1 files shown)


EugenMayer:
That sounds like a fairly good idea - i just recognised why i never implemented that - we are using docker-sync as a library and implemented that part in the library itself.

Nevertheless this makes a lot of sense. I will review the patch and merge that.

Issue #199, opened 2016-12-12
remove unison:onesided in 0.2.0


EugenMayer:
The strategy does not make sense and shall not be used anyway - remove it

EugenMayer:
removed

PR #224, opened 2017-02-01
Update unison strategy to use -testserver instead of sleep

masterful:
This only targets the `unison` sync strategy. I haven't yet updated the `unison-onesided` strategy as I don't use it. Since I haven't tested that I didn't feel comfortable adding it, but I'm willing to do so if you prefer. (I also noticed you plan on removing it in 0.2.0 - #199 )

This should result in marginally faster start up times for `docker-sync`.

It _may_ address #217 as well ¯\\\_(ツ)\_/¯

no sleep


EugenMayer:
unison-onesided will be removed with 0.2.0 so no need to further work on it, its nearly never used.

Thank you for this work once again. I remember @mickaelperrin talked about testserver, i just cannot recall what his point was. @mickaelperrin was there a case>

EugenMayer:
@masterful what do you think about https://github.com/EugenMayer/docker-sync/pull/65#issuecomment-278565262 for logs/pids and management?

PR #225, opened 2017-02-08
support passing & auto-translating prefer options to unison

ignatiusreza:
valid value for `-prefer` are one of the `root` (in our case `src` or `dest`) or one of the special values of `newer` and `older`

while it is possible to pass in `-prefer` to unison via `sync_args`, since `root` are absolute path, it is desirable for it to be automatically translated from either `src` or `dest`, to make it reusable by different member of the team, since people will have different absolute path for the same project

### DIFF
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -121,6 +121,7 @@ module Docker_Sync
         args.push(@options['src'])
         args.push('-auto')
         args.push('-batch')
+        args.push(sync_prefer) if @options.key?('sync_prefer')
         args.push(@options['sync_args']) if @options.key?('sync_args')
         sync_host_port = get_host_port(get_container_name, UNISON_CONTAINER_PORT)
         args.push("socket://#{@options['sync_host_ip']}:#{sync_host_port}")
@@ -129,6 +130,14 @@ module Docker_Sync
           raise('Unison does not support sync_group, sync_groupid - please use rsync if you need that')
         end
         return args
+      end
+
+      def sync_prefer
+        case @options['sync_prefer']
+        when 'src' then "-prefer #{@options['src']}"
+        when 'dest' then "-prefer #{@options['dest']}"
+        else "-prefer #{@options['sync_prefer']}"
+        end
       end
 
       def start_container


EugenMayer:
Could you please elaborate what this is about? -prefer with a src/dest is new to me, @mickaelperrin could you also join this.

ignatiusreza:
sure! to quote on unison documentation:

> prefer xxx
> 
> Including the preference -prefer `root` causes Unison always to resolve conflicts in favor of `root`, rather than asking for guidance from the user. (The syntax of `root` is the same as for the `root` preference, plus the special values `newer` and `older`.) 

which `root` in docker-sync case would contain the expanded form of `src` and `desc`..

in master branch, we already able to tell unison to prefer src by doing something like:

```
syncs:
  something:
    src: '.'
    dest: '/something'
    sync_strategy: 'unison'
    sync_args: [ '-prefer /absolute/path/to/src' ]
```

which would tell unison that in case of conflict, always prefer the version that are on the host machine.. but this are not portable, given that the absolute path will be different, across different machine..

with this PR, we will be able to do something like:

```
syncs:
  something:
    src: '.'
    dest: '/something'
    sync_strategy: 'unison'
    sync_prefer: 'src'
```

EugenMayer:
@ignatiusreza no i get it, sure that make sense. It basically makes the src path potable.

@mickaelperrin any vetos here?

EugenMayer:
@ignatiusreza thank you for this great contribution. Could you give this a try in 0.2.0 and test your feature here? https://github.com/EugenMayer/docker-sync/wiki/How-to-install-the-new-0.2.0-prior-the-release 

Does it work out?

masterful:
@EugenMayer - would it be worth using the `.pre` approach and publishing a pre-release version?
http://help.rubygems.org/discussions/questions/63-how-to-createrelease-a-beta-version-of-a-gem

EugenMayer:
for now, that would be complicated due to the auto-upgrade feeature i have build in, it would not respect pre and upgrade the gem .. that would be not what we want. We wold need to bring out 0.1.6 to fix that, and i do not really want to focus on that

masterful:
Hmm - I'm not sure that's true, though?
If we look at the way gem_update_checker works, it uses the rubygems.org api to see what the latest version is, right? So, for example, rspec has a few pre-release versions listed on their page (the latest being 3.6.0.beta2):
https://rubygems.org/gems/rspec
and if you look at the url you only see the latest stable release (3.5.0):
https://rubygems.org/api/v1/versions/rspec/latest.json

EugenMayer:
hmm thats an interesting point actually, lets move this dicussion to a seperate issue then?

EugenMayer:
@masterful released bete5 https://rubygems.org/gems/docker-sync and latest version is still 0.1.5. Thank you for the suggestion

ignatiusreza:
Thanks guys! I had some difficulties in upgrading from 0.1.5 to 0.2.0, since I'm running ubuntu machine (similar to #248).. but the feature for prefer itself seems to be working! though I noticed something off, and opened a PR at #251

PR #239, opened 2017-02-15
docker-sync-daemon stub (resolves #235)

### LINKED ISSUE #235: create docker-sync-daemon bin-stub
Beside we will offer `docker-sync -d` we could also offer docker-sync-daemon as a stub enforcing this paramter. 

Any thoughts on that @mickaelperrin @masterful ?

masterful:
Resolves #235 

I wasn't sure of a better way to do this, as I'm honestly not that familiar with Thor.

If there's a simpler way I'd love to hear it - I didn't see an easy way to ensure options were passed through without redefining them 😕


EugenMayer:
Actually no, thats not how it works. "Daemon" would need to be a Thor Class, which is not (Only Stack/Sync).

That stub will actually be a bit complicated. Let me help you with that

PR #241, opened 2017-02-16
Add docker-sync log task

DocX:
Motivation: When I run `docker-sync start -d` in background, it is useful to sometimes look at the logs as unison sometimes fails some sync etc.

Benefits:
* Calling `docker-sync log` is more convenient than always manually calling `tail -n 100 .docker-sync/daemon.log`. 
* It hides the implementation of how the daemon works, so if it is subject to change in future, we can keep the same CLI user interface for getting logs.

Task options:
* Duplicated options from `start` - `--dir` and `--app-name`.
* `--lines=N` specifies number of lines to tail
* `-f | --follow` will stream the log

### DIFF
--- a/tasks/sync/sync.thor
+++ b/tasks/sync/sync.thor
@@ -98,6 +98,14 @@ class Sync < Thor
     end
   end
 
+  desc 'log', 'Prints last 100 lines of daemon log. Only for use with docker-sync started in background.'
+  method_option :lines, :aliases => '--lines', :default => 100, :type => :numeric, :desc => "Specify number of lines to tail"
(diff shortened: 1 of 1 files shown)


EugenMayer:
That is actually a very good idea, really nice contribution!

@masterful how is your though about that? i really like it

EugenMayer:
@DocX thanks for taking your time for this, this will become very valueable.

@masterful never forget the docs :) https://github.com/EugenMayer/docker-sync/wiki/2.3-daemon-mode

masterful:
👍  updated the docs

PR #251, opened 2017-02-20
fix checking for 'default' as value for sync_prefer

ignatiusreza:


ignatiusreza:
that's correct.. `-prefer default` is not a valid unison arguments.. it should have been translated into `-prefer [expanded form of src] -copyonconflict`..

masterful:
👌 cool
I'm not sure if @EugenMayer has a procedure he follows (like bumping and publishing version) when merging PRs, so I'm going to wait for him to weigh in on this one first (sorry - was only just added to the contributors, so I'm not familiar with that side of things yet)

ignatiusreza:
not a problem :grin: 

congratulations btw! :tada:

EugenMayer:
Iam not sure what issue we are fixing here and hwo critical it is. Am i missing something, seems to be just more elegant, but .. else?

ignatiusreza:
the problem is, the following code `unless @options.key?('sync_prefer') || @options['sync_prefer'] == 'default'` is not doing what it is expected to do.. when `@options['sync_prefer'] == 'default'` return true, we will fall out of unless, and eventually go into `else "-prefer #{@options['sync_prefer']}"`..

the proper condition to be used are `if !@options.key?('sync_prefer') || @options['sync_prefer'] == 'default'`, which if we want to write it in unless form becomes `unless @options.key?('sync_prefer') && @options['sync_prefer'] != 'default'`.. (`if !A || B` => `if !(A && !B)` => `unless A && !B`)

EugenMayer:
goodness.. well yeah, even looking the second time i overlooked that one! Thank you for the patch

PR #261, opened 2017-02-22
Print version even with no command (fixes #255)

### LINKED ISSUE #255: -v / --version flag not working without subcommand
The `-v`/`--version` flag does not work when used without a subcommand.

```
❯ docker-sync -v
Could not find command "_v".
❯ docker-sync --version
Could not find command "__version".
```

It only works when used with a subcommand.

```
❯ docker-sync start -v                                                                                                         
0.2.0
```

masterful:
The downside to this approach (leaving it in the other commands as well) is that it will print out twice in the list when you get help. I didn't see an easy way to have an option that's hidden, or a command that's undocumented - so if you know of a way I'll happily change it.

This should resolve #255


EugenMayer:
That's neat, did not know about that, thanks!

PR #262, opened 2017-02-23
Show a warning rather than an error

masterful:
If the sync is already running, then it may look nicer to show a warning message rather than a red error message. That way if someone has it as part of a start-up script that they end up running more than once they'll just get a yellow warning message rather than a worrisome red error.
(a few of my coworkers noticed this, which means it's likely others will as well)


EugenMayer:
Makes a lot of sense, nowadays with the daemon mode even more. Thanks!

Issue #267, opened 2017-03-07
ability to configure sync manager with yaml string instead of config_path


eelkevdbos:
The ability to supply a (yaml formatted) config string is a welcome addition when composing your own sync script. It prevents writing a file to disk and makes dynamic configuration a more viable strategy.

**Proposal**: 
- Do not raise an exception if an inline configuration was supplied during initialisation
- Add a method to the ConfigTemplate class that will 'just' YAML load the config_string supplied

EugenMayer:
Great addition, thank you for taking care!

PR #268, opened 2017-03-07
Feature/inline configuration

eelkevdbos:
Implementing feature #267


EugenMayer:
Sorry for beeing late. Eventhough i do not like the current configuration loader at all, one of those reasons is the one you partially fixed now, i see the value in adding your path now, its one step further into the right direction.

Thank you a lot for your contribution!

Issue #256, opened 2017-02-21
Make dest in sync-configuration optional


matleh:
Version 0.2 uses volumes (way to go!) which are much more flexible than the old data-containers.
As I understand it, the `dest` configuration parameter is no longer needed, since the mount-point is now given when the volume is used. The sync-container is just there to copy the data between host and volume - it can mount the volume at any location, without consequences for the container that uses the volume.

However, specifying a sync without `dest` throws an error during `docker-sync start`.

```
/usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:95:in `block in validate_sync_config': fuse_src does not have dest configuration value set - this is mandatory (RuntimeError)
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:94:in `each'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:94:in `validate_sync_config'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:85:in `block in validate_config'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:84:in `each'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:84:in `validate_config'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:35:in `load_configuration'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/lib/docker-sync/sync_manager.rb:25:in `initialize'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/tasks/sync/sync.thor:30:in `new'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/docker-sync-0.2.0/tasks/sync/sync.thor:30:in `start'
	from /usr/local/Cellar/gem-docker-sync/0.2.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
[...]

PR #271, opened 2017-03-13
Make 'dest' optional in sync configuration

matleh:
If not given, use default value '/sync'.

This is a very minimal change for #256 . I tried it with unison - not sure about rsync since the trailing '/'  has special meaning there.


EugenMayer:
once again, great contribution!  Thank you very much

Issue #254, opened 2017-02-20
Configure reading `.env` files


sloria:
#230 introduced support for reading `.env` files and using envvards in docker-sync configuration.

I use [autoenv](https://github.com/kennethreitz/autoenv) in order to do more than just set environment variables in my `.env` file (setting my Python virtual environment, for example). Therefore, I get an error when I run `docker-sync start`:

```
/usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv/parser.rb:63:in `parse_line': Line "source activate osf" doesn't match format (Dotenv::FormatError)
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv/parser.rb:47:in `block in call'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv/parser.rb:46:in `each'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv/parser.rb:46:in `call'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv/parser.rb:36:in `call'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv/environment.rb:13:in `load'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv/environment.rb:9:in `initialize'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv.rb:15:in `new'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv.rb:15:in `block (2 levels) in load'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv.rb:59:in `ignoring_nonexistent_files'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv.rb:14:in `block in load'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv.rb:46:in `block in with'
        from /usr/local/lib/ruby/gems/2.4.0/gems/dotenv-2.2.0/lib/dotenv.rb:45:in `each'
[...]

PR #274, opened 2017-03-17
make name of env-file configureable

matleh:
Use DOCKER_SYNC_ENV_FILE to set filename of env-file.
Defaults to ".env"

This takes care of #254


EugenMayer:
Great stuff, thank you a lot for implementing this!

PR #290, opened 2017-04-12
Updating precondition checks for unison-fsmonitor compatibility

MMartyn:
unison-fsmonitor has moved to using watchdog instead of macfsevents which is causing new installs of docker-sync to fail to start.

Relevant commit:
https://github.com/hnsl/unox/commit/2c60f8e0436a93ab05d61bebf09d48309c4eab6f

### DIFF
--- a/lib/docker-sync/preconditions.rb
+++ b/lib/docker-sync/preconditions.rb
@@ -6,7 +6,7 @@ module Preconditions
     docker_running
     unison_available
     unox_available
-    macfsevents_available
+    watchdog_available
   end
 
   def self.docker_available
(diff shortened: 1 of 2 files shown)


EugenMayer:
That is a pretty interesting point. Unox is the filewatch on OSX, @mickaelperrin did you notice that?
@MMartyn newer installation basically install ah newer unox and therefore having a watchdog dependency - thats your point right?

What if someone upgrades docker-sync to the upcoming version, but has the old unox installed, it wont get replaced, thus we would handle the wrong dep.

To make this PR work we need to define a minimum version on unox to match the one requiring watchdog.

Other then that, thank you for the work, lets just finish this. I guess i will have to release a new version out of this!

MMartyn:
Yeah, that was the issue. Old version requires macfsevents and the new version requires watchdog. Unfortunately, unox isn't versioned, so it's unclear how you would go about implementing a version check.

malagant:
Great. I just found the solution an hour ago and wanted to create a pull request. Glad you already did it. :)

EugenMayer:
So what you guys think, with the potential to break other setups, how shall we include this one. Its kinda risky, right?

malagant:
Well. My setup started to work with just installing watchdog using pip.
Maybe updating the wiki with a hint about watchdog is enough.

EugenMayer:
What about simply install both for now, lets say until 2.1.5?

malagant:
I prepared a solution in #295 . You are free to merge or not ;)

MMartyn:
Going to close this as I see you merged #295

PR #291, opened 2017-04-13
add `max_attempt` options

donnykurnia:
to set how many `attempt` to wait for the server up and running before failed.
In my case, 5 attempts often not enough, so I want to have it configurable via options.

### DIFF
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -184,11 +184,12 @@ module Docker_Sync
         cmd = "unison -testserver #{@options['dest']} \"socket://#{@options['sync_host_ip']}:#{sync_host_port}\""
         say_status 'command', cmd, :white if @options['verbose']
         attempt = 0
+        max_attempt = @options['max_attempt'] || 5
         loop do
           stdout, stderr, exit_status = Open3.capture3(cmd)
           break if exit_status == 0
           attempt += 1
-          break if attempt > 5
+          break if attempt > max_attempt
           sleep 1
         end
         sync


EugenMayer:
You could easily use sync_args for this, right?

donnykurnia:
@EugenMayer `sync_args` is passed to unison and it's used by unison instead of the `docker-sync` tasks. I have tried to add `sync_args: [ '-retry 5' ]` and when I look up the running process, it's passed to unison correctly:

```
ps A|grep -i unison
65578   ??  S      0:00.87 unison -ignore=Name .DS_Store -ignore=Name ._* /Users/donnykurnia/data/Projects/project_name -auto -batch -prefer /Users/donnykurnia/data/Projects/project_name -copyonconflict -retry 5 socket://192.168.101.100:32770 -repeat watch
```

In my use-case, however, the docker-machine sometimes didn't start the docker container in time and the `docker-sync` task reaches 5 attempts and quit. I had to run the `docker-sync-daemon start` again to and it will work because the docker container already started and the `docker-sync` only restart the container which most of the time faster than starting a stopped container.

Right now I changed the `max_attempt` to 10 and for the past 2 days, the docker-sync-daemon start always successfully. This might not for everybody, but it's best to have this configured via `options` instead of having to edit the ruby file directly.

EugenMayer:
@donnykurnia thank you for clarifying, i missed the point beforehand. 

Please add the option to https://github.com/EugenMayer/docker-sync/blob/master/example/docker-sync.yml and explain it, and also update https://github.com/EugenMayer/docker-sync/wiki/2.-Configuration

so we keep the docs up2date - since that matters :) After that i happily merge this one.

In case i might help you also another way - you most probably running into fsync issues, macing d4m even slower.

Try running this:
https://gist.github.com/EugenMayer/55fe8c7a3e8c27ecf6a9a52f58a64be8

This will speed up d4m about 10 times on usual filesystem speed ( not the synced folders, the normal fs performance is broken right now ) .. and that will be the reason for the slowness you experience.

If that helps you, please open a ticket, we might consider communicating this to the docker-sync people and also provide either tooling or at least doc hints to do that

EugenMayer:
i am going to release 2.1.4 today, if you like, try to finish this, so we can pick this one up

EugenMayer:
Planing to decline this PR after 7 days if no progress is made, since i expect this effort to be no longer interesting for the author and probably no longer needed

donnykurnia:
@EugenMayer I just check this PR today. FYI, I'm using docker-machine in my MacBook (using VirtualBox), not docker for mac.

I'll add the option and explanation now.

EugenMayer:
thank you for your great work!

PR #293, opened 2017-04-13
Make docker-sync-daemon stop synchronously

michaelbaudino:
## Problem

When running `docker-sync-daemon stop`, we send a system signal (`INT`) to the daemonized process but we return immediately (_i.e._ before the daemonized process has actually stopped).

This is inconsistent with all the efforts put in making `docker-sync-daemon start` synchronous (it returns only when the daemon is started and the initial sync has completed).

## Solution

It's a 2-parts solution:

1. Use [`docker wait`](https://docs.docker.com/engine/reference/commandline/wait/) to ensure the unison Docker container has actually stopped.
2. After sending the `INT` signal to the daemonized process group, watch the output of `ps` until the PID is gone.

## Exemple

Before this PR (when `docker-sync-daemon` is started, indeed):

```shell
$ docker-sync-daemon stop
    shutdown  Background docker-sync has been stopped
$ docker-sync-daemon stop
    shutdown  Background docker-sync has been stopped
$ docker-sync-daemon stop
    shutdown  Background docker-sync has been stopped
$ docker-sync-daemon stop
       error  No such file or directory - ./.docker-sync/kyero.pid
       error  Check if your PIDFILE and process with such PID exists
```

After this PR:

```shell
$ docker-sync-daemon stop
    shutdown  Waiting for background docker-sync to terminate
    shutdown  Background docker-sync has been stopped
$ docker-sync-daemon stop
       error  No such file or directory - ./.docker-sync/kyero.pid
       error  Check if your PIDFILE and process with such PID exists
```


EugenMayer:
That's a really good catch also. We should though use a maximum wait timeout, lets say 30 seconds, then we bail out telling that we could be it end it, what do you think?

michaelbaudino:
Good point 👍 Since there's a bit more logic, I did it in Ruby ♦️♦️♦️ (rather than previous version in shell script)

EugenMayer:
@michaelbaudino thanks! could i encourage you to use this kind of implementation?

    puts 'Waiting for process to go done...'
        Timeout::timeout(30) do
          loop do
            sleep(2)
            print '.'
            return true if check_process_is_down
            next
          end
        end

So basically Timeout::timeout

michaelbaudino:
Here you go 👍 

The `puts` on one line are a bit ugly in code (specially because they force us to use multiline `if...then...else...` clauses), but at least they make the CLI output clean. I can change it the other way around if you prefer, like:

```ruby
def wait_for_process_termination(pid)
  print 'Waiting for background docker-sync to terminate'
  Timeout::timeout(30) do
    loop do
      return say_status 'shutdown', 'Background docker-sync has been stopped' if process_dead?(pid)
      sleep 1
      print '.'
    end
  end
rescue Timeout::Error
  say_status 'error', 'Background docker-sync daemon failed to stop within 30 seconds', :red
  exit 70 # EX_SOFTWARE (according to `man sysexits`)
end
```
(code cleaner, but CLI output messed up because some `\n` are lacking)

EugenMayer:
please make the pull request compatible with the current master branch, thank you!

michaelbaudino:
Done ✔️

PR #294, opened 2017-04-13
Return an error exit code when trying to stop a daemon which is not running

michaelbaudino:
## Problem

When `docker-sync-daemon` is not started, running `docker-sync-daemon stop` outputs an error message but exits with a success error message (`0`, _aka_ `EX_OK`).

## Solution

According to `man sysexits` (which documents `/usr/include/sysexits.h` on BSD systems), we might want to exit with error code `69` (_aka_ `EX_UNAVAILABLE`):

> **EX_UNAVAILABLE (69)**   A service is unavailable.  This can occur if a support program or file does not exist.  This can also be used as a catchall message when something you wanted to do doesn't work, but you don't know why.

## Exemple

Before this PR (without `docker-sync-daemon` started):

```shell
$ docker-sync-daemon stop
       error  No such file or directory - ./.docker-sync/kyero.pid
       error  Check if your PIDFILE and process with such PID exists
$ echo $?
0
```

After this PR:

```shell
$ docker-sync-daemon stop
       error  No such file or directory - ./.docker-sync/kyero.pid
       error  Check if your PIDFILE and process with such PID exists
$ echo $?
69
```

### DIFF
--- a/tasks/sync/sync.thor
+++ b/tasks/sync/sync.thor
@@ -61,9 +61,8 @@ class Sync < Thor
       say_status 'shutdown', 'Background dsync has been stopped'
     rescue Errno::ESRCH, Errno::ENOENT => e
       say_status 'error', e.message, :red # Rescue incase PIDFILE does not exist or there is no process with such PID
-      say_status(
-        'error', 'Check if your PIDFILE and process with such PID exists', :red
-      )
+      say_status 'error', 'Check if your PIDFILE and process with such PID exists', :red
+      exit(69) # EX_UNAVAILABLE (see `man sysexits` or `/usr/include/sysexits.h`)
     end
   end
 


EugenMayer:
Good find! Are you using docker sync in an automated CI CD tool? Thank you for the contribution!

michaelbaudino:
Not in CI/CD, but I provide a wrapper script around `docker-sync-daemon` to my team and I realized that my wrapper was saying "Success" even when running `docker-sync-daemon stop` when no daemon was running (because my wrapper was checking the exit status), which was weird.

Thank **you** for your work 😊

EugenMayer:
@michaelbaudino are you aware of the scaffolding mechanism?  https://github.com/EugenMayer/docker-sync/wiki/7.-Scripting-with-docker-sync

Telling, since i do also provide a wrapper script, but in ruby + thor, so basically something liek "start project" which does not only start the docker stack, but also triggers a example db installation or stages a customer + does dev settings on the installation like switching of aggregations and stuff.

We do also user docker-sync start + daemon, but also in full custom mode, so starting stync-start + docker-compose manually no bash scripts.

If you really want to orchestrate docker-sync, i yet found the scaffolding / lib way to be very usefull - just sharing that, maybe that helps you

PR #295, opened 2017-04-13
Refactored pip install method and added watchdog installation to fix …

malagant:
…problem on Mac OS X


VERSION
```diff
@@ -1 +1 @@
-0.2.3
+0.2.4
```
EugenMayer: remove this please, no version bumps

EugenMayer:
That is really good, well done - thank you for the contribution. Could you please remove the version bump and then we are good to go!

EugenMayer:
Great work and huge thanks taking your time to contribute.

Issue #265, opened 2017-03-03
Make docker-sync cross-platform


ignatiusreza:
### Goals

Summarizing the discussion in #250, there are 2 goals in general that we want to achieve in order to make docker-sync cross-platform:

- make platform & strategy specific codes pluggable
- add `volume` strategy which would use `docker volume create` natively

### ~~Linux Native Support~~

~~Since docker have nice linux support, docker-sync are basically not needed under linux.. but, in order to allow users to share the same docker-compose file between osx and linux users, we would need to be able to somehow parse the config files, and then boot up docker-compose with a [modified docker-compose configuration](https://github.com/xeger/docker-compose/issues/27).. for example, the following config:~~

```yml
# docker-compose.yml
version: "2"
services:
  app:
    volumes:
      - syncer:/var/www

volumes:
  syncer:
    external: true

# docker-sync.yml
version: '2'
syncs:
  syncer:
    src: './data'
    dest: '/var/www'
```

~~would be modified into~~

```yml
# docker-compose.yml
version: "2"
services:
  app:
    volumes:
      - ./data:/var/www
```

~~first before being used to boot up docker-compose~~

### Cross Platform Pluggable Strategies

To make our code base cross platform, we would need to restructure code related to configuration and dependency (rspec, fswatch & unison) management into something like:

- `DockerSync::Config`: responsible for parsing the configuration files and turn it into PORO

  instance of this class will hold array of instances of either `DockerSync::Config::Rsync`, `DockerSync::Config::Unison` under `#services`.. each service provider would need to handle parsing raw configuration from yml into a format that is understandable by the actual sync provider (to be pass in as command line arguments)..

- `DockerSync::Dependency`: responsible for checking availability and auto installing dependencies.. it will be read from `DockerSync::Config` and then run platform specific code to check and auto install as necessary.. the code can be structured like:

  - `DockerSync::Dependency`:

    - `DockerSync::Dependency::OSX`:

      - `DockerSync::Dependency::OSX::Fswatch`
      - `DockerSync::Dependency::OSX::Rsync`
      - `DockerSync::Dependency::OSX::Unison`

    - `DockerSync::Dependency::Volume`

    - etc

  with the possibility for registering more dependencies by doing something like `DockerSync::Dependency.register_dependency SomeClass, ->(config) { config.sync_strategy == 'awesome_strategy' }`..

[...]

PR #301, opened 2017-04-24
implemented precondition os based strategy, implemented brew being ma…

EugenMayer:
implemented precondition os based strategy, implemented brew being ma…ndatory, prepared 0.3.0 release, fixed upgrade check being run on first installation. Install osx libs using brew only

@ignatiusreza have a look at this PR, its is most probably gearing towards what you do or started.

i implemented this since i want to make brew mandatory on OSX, this would have broken Linux support forever, so i implemented it the proper way as we planned.

Linux is yet stubbed as you see - you dare tackling this? :)


ignatiusreza:
@EugenMayer WOW! That's quite some changes! I'll see what I can do on the Linux branch..

before that though.. let's clarify on when we want to on support for linux.. in #265 we agreed that unison/rsync shouldn't be needed in linux, and instead we could introduce a new sync strategy using native `docker volume create`.. after that, we can update [this normalization steps](https://github.com/EugenMayer/docker-sync/blob/0.3.0/lib/docker-sync/config/project_config.rb#L101-L120) into something like:

```rb
def sync_strategy_for(sync_config)
  sync_strategy = sync_config['sync_strategy']

  if ['rsync', 'unison', 'volume'].include?(sync_strategy)
    sync_strategy
  else 
    default_sync_strategy
  end
end

def watch_strategy_for(sync_config)
  watch_strategy = sync_config['watch_strategy']
  watch_strategy = 'dummy' if watch_strategy == 'disable'

  if ['fswatch', 'unison', 'dummy'].include?(watch_strategy)
    watch_strategy
  else
    default_watch_strategy
  end
end

def default_sync_strategy
  case true
  when OS.linux? then 'volume'
  else 'unison'
  end
end

def default_watch_strategy
  case sync_strategy_for(sync_config)
  when 'rsync' then 'fswatch'
  when 'unison' then 'unison'
  when 'volume' then 'dummy'
  else raise "you shouldn't be here"
  end
end
```

which would be suffice in making sure that docker-sync is usable in linux.. we can also add some check under `preconditions_linux` that the necessary binary is available if user are so inclined on using `rsync` or `unison`, but I'm not so sure in auto installing unison binary.. wdyt?

PR #307, opened 2017-04-26
bugfix on missing config_path when using fswatch as watch_strategy

ignatiusreza:
2bf2a6d0348f85e90924821b4b0e9940c0372292 fix it by removing support for `config_path`, but the root issue is we forgot to set `@config_path` in `SyncManager`, which are used in [this line](https://github.com/EugenMayer/docker-sync/blob/be86d1bf546bf964d8ba6970af51caf5ac558db2/lib/docker-sync/sync_manager.rb#L31)

so in this PR, I brought back support for `config_path`, and fix fswatch by properly setting `@config_path` (+ some extras :grin: )...

### DIFF
--- a/lib/docker-sync/config/project_config.rb
+++ b/lib/docker-sync/config/project_config.rb
@@ -26,7 +26,10 @@ module DockerSync
 
     def initialize(config_path: nil, config_string: nil)
       if config_string.nil?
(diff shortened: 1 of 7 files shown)


EugenMayer:
nice cleanup there, had been too late lately. Very nice work. i will wait for some time to see if we run into further bugs and the make a 0.3.6 with all those things inside!

PR #309, opened 2017-04-26
bugfix: method_missing watch_fork on dummy & fswatch watch strategy

ignatiusreza:
called from: https://github.com/EugenMayer/docker-sync/blob/4766014b8faed574e3153ff244177e8e0fdf5224/lib/docker-sync/sync_process.rb#L90


EugenMayer:
nice!

PR #310, opened 2017-04-26
add sync strategy for native docker volume

ignatiusreza:
to be used by default in linux environment, since docker built in support for mounting volume is good enough and we can use it without relying on rsync or unison..

part of #265 

NOTE: I'm currently using this branch in my local.. give it some time to see if there're any issue, and I'll report back later..

### DIFF
--- a/example/docker-sync.yml
+++ b/example/docker-sync.yml
@@ -23,8 +23,18 @@ options:
   # optional, default: pwd, root directory to be used when transforming sync src into absolute path, accepted values: pwd (current working directory), config_path (the directory where docker-sync.yml is found)
   project_root: 'pwd'
 syncs:
+  default-sync:
+    # os aware sync strategy, default to unison under osx, and native docker volume under linux
+    sync_strategy: 'default'
+    # which folder to watch / sync from - you can use tilde, it will get expanded.
+    # the contents of this directory will be synchronized to the Docker volume with the name of this sync entry ('default-sync' here)
+    src: './default-data/'
+
+    # other unison options can also be specified here, which will be used when run under osx,
+    # and ignored when run under linux
+
   unison-sync:
-    # unison 2 way-sync, default since 0.2.0
+    # unison 2 way-sync
     sync_strategy: 'unison'
     # common options
     # see rsync documentation for all these common options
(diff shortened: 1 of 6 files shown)


EugenMayer:
Hmm i really like the implementation, but i have some concerns i want to share.

Right now, you either chose a strategy explicitly or you get the default ( unison ). with your implementation we have to adjust a lot of docs to reflect that the default is platform specific.

Beside this effort,  it makes really a lot of sense having platform specific defaults , since they have different OS capabilties.

That said, i guess technically, its the right move. But are you opting it for all those doc changes? ( wiki, examples and so on? )

ignatiusreza:
yup, I'll update the examples once I give it some try locally to see if it's actually working.. regarding wiki, what's the timing for updates? I guess it should be updated only once new release are out, so that people who bound into them can actually use them.. and, any other places that need to be updated? noticed that you also have the boilerplate repo, need to be updated?

EugenMayer:
I have an additional very interesting idea here, which could end up being one of the nicest addition since a long time. Your work here can be used for that too, makes it even more important.

Consider that we do the following on osx : native mount (osxfs) [host sync strategy/watcher] <-> sync container (unison) <-> named volume mount <-> app

This way we could remove unison/unox pain under osx, the pain of the native dependencies and also it's flaws while still maintaining the native speed in the app. What do you think?

ignatiusreza:
hmm.. give me some time to digest that first.. :sweat_smile: not that comfortable yet with how sync/watch works..

ignatiusreza:
updated example with new section for default configuration

> Consider that we do the following on osx : native mount (osxfs) [host sync strategy/watcher] <-> sync container (unison) <-> named volume mount <-> app

Regarding this.. I'm still not completely sure what to say to be honest... still trying to wrap my head around how the actual syncing/watching process works and how does it affect the performance.. though, IIUC.. the reason that docker-sync is needed, is because native mount on osx is slow.. so, in the graph above, won't doing `native mount (osxfs) [host sync strategy/watcher] <-> sync container (unison)` be slow? if it's still fast, why does mounting into sync container is okay, but mounting directly into app is not?

EugenMayer:
@ignatiusreza lets continue the new strategy in a the issue and keep this one for linux only for now - lets see if we can merge the efforts here later

ignatiusreza:
Cool! thanks for merging :+1:

PR #312, opened 2017-04-26
Remove unused install_pip method

michaelbaudino:
This is just some sanity cleanup: this method is not used anymore (and will hopefully not ever be again) ✂️ ✂️ ✂️


EugenMayer:
well i doubt it will not, i would for now like to keep it. Lets assume 3 months, if nothing happens, lets remove it, ok?

michaelbaudino:
Gotcha 👍
Closing for now then.

PR #314, opened 2017-04-27
bugfix: src should be expanded relative to where the config is found

### LINKED ISSUE #311: docker-sync creates a directory for each of external volumes inside project root directory
docker-sync creates a directory for each of external volumes inside project root (http://d.pr/i/9wrwu). This affects editor indexing and so on.

Excerpt of docker-compose.yml file

        services:         
              web:
                    container_name: web
                    build:
                        context: .
                        dockerfile: .docker/docker-web.dockerfile
                    volumes:
                        - ezpublish5-sync:/var/www/ezpublish5:rw
                        - ubb-sync:/var/www/www.terrengsykkel.no:rw
                        - ffshop-sync:/var/www/www.ffshop.no:rw
                    working_dir: /var/www/ezpublish5
                    depends_on:
                        - db
                    networks:
                        - app-back
                        - app-front
        volumes:
            ezpublish5-sync:
                external: true
            ffshop-sync:
                external: true
            ubb-sync:
                external: true

docker-sync.yml

        version: "2"
        
        options:
            verbose: true
            # default: docker-compose.yml if you like, you can set a custom location (path) of your compose file like ~/app/compose.yml
            compose-file-path: 'docker-compose-mac.yml'
        
        syncs:
            #IMPORTANT: ensure this name is unique and does not match your other application container name
            ezpublish5-sync: #tip: add -sync and you keep consistent names als a convention
                src: '.'
                dest: '/var/www/ezpublish5'
                sync_excludes: ['.docker']
                watch_excludes: ['.*/.git', '.*/node_modules', '.*/bower_components', '.*/sass-cache', '.*/.sass-cache', '.*/.sass-cache', '.coffee', '.scss', '.sass', '.gitignore']
        
            ubb-sync: #tip: add -sync and you keep consistent names als a convention
                src: '../ubb'
                dest: '/var/www/www.terrengsykkel.no'
                watch_excludes: ['.*/.git', '.*/node_modules', '.*/bower_components', '.*/sass-cache', '.*/.sass-cache', '.*/.sass-cache', '.coffee', '.scss', '.sass', '.gitignore']
        
            ffshop-sync: #tip: add -sync and you keep consistent names als a convention
                src: '../ffshop'
                dest: '/var/www/www.ffshop.no'
                watch_excludes: ['.*/.git', '.*/node_modules', '.*/bower_components', '.*/sass-cache', '.*/.sass-cache', '.*/.sass-cache', '.coffee', '.scss', '.sass', '.gitignore']

ignatiusreza:
instead of pwd, so that it will get resolved deterministically when run in child directories..

fix #311

### DIFF
--- a/example/docker-sync.yml
+++ b/example/docker-sync.yml
@@ -20,6 +20,8 @@ options:
   cli_mode: 'auto'
   # optional, maximum number of attempts for unison waiting for the success exit status. The default is 5 attempts (1-second sleep for each attempt). Only used in unison.
   max_attempt: 5
+  # optional, default: pwd, root directory to be used when transforming sync src into absolute path, accepted values: pwd (current working directory), config_path (the directory where docker-sync.yml is found)
+  project_root: 'pwd'
 syncs:
   unison-sync:
     # unison 2 way-sync, default since 0.2.0
--- a/lib/docker-sync/config/project_config.rb
+++ b/lib/docker-sync/config/project_config.rb
@@ -90,16 +90,26 @@ module DockerSync
       end
 
       def normalize_config!
+        normalize_options_config!
+
         config['syncs'].each do |name, sync_config|
           config['syncs'][name] = normalize_sync_config(sync_config)
         end
+      end
+
+      def normalize_options_config!
+        config['options'] = {
+          'project_root' => 'pwd',
+        }.merge(config['options'] || {})
       end
 
       def normalize_sync_config(sync_config)
         {
           'sync_strategy' => sync_strategy_for(sync_config),
           'watch_strategy' => watch_strategy_for(sync_config)
-        }.merge(sync_config)
+        }.merge(sync_config).merge(
+          'src' => expand_path(sync_config['src']),
+        )
       end
 
       def sync_strategy_for(sync_config)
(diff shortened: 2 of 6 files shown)


EugenMayer:
We should be carefully with this change, it could break setups. It's not a bad idea to have it relative to the config, but we should make this configureable for now, default being pwd...

We could think about changing the default with 0.4 later, since relative to the config might be the best ootb way.

I would merge this now, but only with the lagacy default + configureable. So we need doc updates and config in this pr

ignatiusreza:
agreed that it is a breaking change.. if we make it configurable, should it be configurable on the sync level, or project level? I think it makes more sense to have it under `options` (project level), but some people might want to have more find grain control..

EugenMayer:
i would even tend to talk about doing that into the gloabl options, but at maximum on project options, not sync options ( thats goes way to far )

ignatiusreza:
project level option works better I guess.. since it ties into the project `docker-sync.yml`.. some people might be using docker-sync in multiple projects, and not have the room to upgrade the settings for all of them at the same time.. and with rvm/rbenv, it's possible to have different version of docker-sync on different projects..

ignatiusreza:
updated to be configurable under project options

EugenMayer:
Very nice, we need to also update the wiki configuration page and that's it. Once again, well done!

ignatiusreza:
wiki updated :+1:

Issue #320, opened 2017-04-28
Error starting sync, exit code 0


VSilantyev:
Every time when I first start docker-sync after reboot, it gives me this error.
Sync is not working after it.

When I stop this and run again - it will be successful.

Logs for cold start: 
```bash
me@mybook:/www/myproject$ docker-sync-stack start
       note:  You can also run docker-sync in the background with docker-sync --daemon
          ok  Starting unison
          ok  myproject-unison-sync container not running
          ok  starting myproject-unison-sync container
     command  docker start myproject-unison-sync && docker exec myproject-unison-sync supervisorctl restart unison
          ok  starting initial sync of myproject-unison-sync
     command  docker inspect --format='{{(index (index .NetworkSettings.Ports "5000/tcp") 0).HostPort}}' myproject-unison-sync
     command  unison -testserver /sync "socket://127.0.0.1:32768"
     command  docker inspect --format='{{(index (index .NetworkSettings.Ports "5000/tcp") 0).HostPort}}' myproject-unison-sync
     command  unison -ignore='Path .git' -ignore='Name .ssh' -ignore='Name .idea' -ignore='Name app/cache/dev/profiler' -ignore='Name app/cache/dev/jms_diextra' -ignore='Name app/cache/codeception/jms_diextra' -ignore='Name app/cache/dev/assetic' -ignore='Name app/cache/test/assetic' -ignore='Name app/cache/test/profiler' -ignore='Name app/cache/dev/annotations' -ignore='Name app/cache/codeception/annotations' -ignore='Name app/cache/test/annotations' -ignore='Name app/cache/codeception/jms_security' -ignore='Name app/cache/staging/jms_security' -ignore='Name app/cache/dev/jms_security' -ignore='Name app/cache/codeception/sessions' -ignore='Name docker-sync.yml' /www/myproject -auto -batch -prefer '/www/myproject' -copyonconflict socket://127.0.0.1:32768
       error  Error starting sync, exit code 0
     message  
     message  Contacting server...
Fatal error: Lost connection with the server
     success  Unison server started
     command  docker inspect --format='{{(index (index .NetworkSettings.Ports "5000/tcp") 0).HostPort}}' myproject-unison-sync
[...]

PR #325, opened 2017-05-01
raise error instead of failing silently when unison failed to start in time

ignatiusreza:
related to: #217, #291, #320, and several others

even though it doesn't really solve the issue of why the first run is slow, it would help user to understand what they need to do in order to overcome the issue.. and I find it odd to break out the loop silently in both success and failure case.. failure case should be loud so that we know when it does failed..


EugenMayer:
very good point!

PR #327, opened 2017-05-02
Delete idea folder

Saphyel:
This folder shouldn't be there.


EugenMayer:
And why so? Why should I consider reconfigure the project each checkout?

rodrigoaguilera:
http://stackoverflow.com/questions/31764319/should-i-commit-my-idea-folder

EugenMayer:
As long as maintainance and development happens through the hands of very few and those few use the same IDE ( guess out of reasons ) it a very good tradeoff to commit it. Some files should be excluded, be it workspace.xml or likes, but most of the things do actually make a lot of sense.

I am not a purist - i do what works. And that works - sorry.

Saphyel:
@EugenMayer then can you explain this line? https://github.com/EugenMayer/docker-sync/blob/master/.gitignore#L1

EugenMayer:
@Saphyel it a gitignore template, was not intended, i will fix it, thanks for pointing me on that

Saphyel:
@EugenMayer np, glad to help even if is in the wrong way ^^

EugenMayer:
If you know what to share, its the best option, even for larger teams. You have to use deploy server for remote SDK and things, to avoid static stuff, but beside some anoyances with ruby-gem changes, its very handy, getting all the run configs, test setups and project setting - sure in a company env, where people have the same ruby SDK installed, e.g. deployed by boxen and things.

In OSS projects, i always check those in since mainly, i do 99% of the work anyway. It becomes different for docker-sync slowly and i will go a different route, if enough demand and good arguments are around. Feel free to start a discussion

PR #338, opened 2017-05-05
Fix typos in deprecations warnings...

michaelbaudino:
... daemon mode **is** the default 😄


EugenMayer:
it obviously was late.. :) thanks

Issue #357, opened 2017-05-06
`docker-sync-stack start` executes `docker-compose up` before initial sync is completed


rwilliams:
### Error
`docker-sync-stack start` executes `docker-compose up` before the initial sync is completed

### Docker Driver
d4m

### Sync strategy
native_osx 
docker-sync 0.4.2

### your docker-sync.yml
```
version: "2"

options:
  verbose: true

syncs:
  myappname-sync: 
    sync_excludes: ['.gitignore', '.idea/*','.git/*', '*.coffee', '*.scss', '*.sass','*.log']
    src: './'
```

### OS
OSX 10.12.4

EugenMayer:
That is impossible to fix right now, i would assume. Do not expect this to get fixed quick or fast - though i understand the need of it

rwilliams:
Yep, no worries. Now that we have it documented it'll be super easy for folks to work around.

EugenMayer:
should have been solved by #366 in the unison image

EugenMayer:
try

```
gem uninstall docker-sync 
gem install docker-sync --version 0.4.3-beta1
```

Does it help with the synchronity?

EugenMayer:
i should have fixed this now with

```
gem uninstall docker-sync
gem install docker-sync --version 0.4.4.beta1
```

Please report back

PR #362, opened 2017-05-07
Wait for Unison to start before tailing its log

michaelbaudino:
This PR fixes the error when unison does not start fast enough and that we try to `tail` on the log file that has not yet been created.

```
   Sync Log:  tail: can't open '/tmp/unison.log': No such file or directory
   Sync Log:  tail: no files
```

ℹ️ to easily reproduce, I run `docker-sync clean` before running `docker-sync start`

### DIFF
--- a/lib/docker-sync/watch_strategy/remotelogs.rb
+++ b/lib/docker-sync/watch_strategy/remotelogs.rb
@@ -22,7 +22,8 @@ module Docker_Sync
 
       def run
         say_status 'success', "Showing unison logs from your sync container: #{@unison.get_container_name}", :green
-        cmd = "docker exec #{@unison.get_container_name} tail -f /tmp/unison.log"
+        cmd = "docker exec #{@unison.get_container_name} tail -F /tmp/unison.log"
+        wait_for_unison_log
         @watch_thread = threadexec(cmd, 'Sync Log:')
       end
 
@@ -44,6 +45,26 @@ module Docker_Sync
 
       def watch_thread
         return @watch_thread
+      end
+
+      private
+
+      def wait_for_unison_log
+        say_status 'wait', "Waiting for unison to start", :white
+        Timeout::timeout(30) do
+          loop do
+            return if unison_log_available?
+            sleep 1
+          end
+        end
+      rescue Timeout::Error
+        puts
+        say_status 'error', 'Unison failed to start within 30 seconds.', :red
+        exit 70 # EX_SOFTWARE (according to `man sysexits`)
+      end
+
+      def unison_log_available?
+        system("docker exec #{@unison.get_container_name} test -f /tmp/unison.log")
       end
     end
   end


EugenMayer:
We should also address this one https://github.com/EugenMayer/docker-sync/issues/357 but i guess, we have to split efforts, right?

EugenMayer:
we might not just wait for the log to spawn, but rather check for a specific string which tells us something like "initial sync done"

EugenMayer:
i solved #366 in the image but also added /tmp/unison.log to the image https://github.com/EugenMayer/docker-image-unison/commit/545bec3d4e47b4e2a991d21901d211f0dcecf0a2 this should make this obsolete and robust, would it?

The point with your wait is, that if the src folder is empty during the initial start, the log is never created and the watch is getting a timeout. Sure, its unpractical but can happen - the way adding it to the image deals with both.

What do you think?

EugenMayer:
try

```
gem uninstall docker-sync 
gem install docker-sync --version 0.4.3-beta1
```

And tell me how you think it works without this patch in your env

michaelbaudino:
Yup: https://github.com/EugenMayer/docker-image-unison/commit/545bec3d4e47b4e2a991d21901d211f0dcecf0a2 should fix it 👍 

But I think we still need to use `tail -F` rather than `tail -f` (because `-F` reloads the file if it is re-created by an external process). Here is an extract from `man tail`:

```
       -f, --follow[={name|descriptor}]
              output appended data as the file grows; -f, --follow, and --follow=descriptor are equivalent

       -F     same as --follow=name --retry

       --retry
              keep trying to open a file even when it is or becomes inaccessible; useful when following by name, i.e., with --follow=name
```

Do you want a PR for this ?

EugenMayer:
That makes a lot of sense actually. 

No jus tcommit that one, no need to overcomplicate thinks.

Thank you for joining the fight - close this PR if you consider it irrelevant

michaelbaudino:
Yup, I pushed  directly to `master` the `tail -F` thing and also a fix of the version string that was causing this error:

```
→ docker-sync start --foreground
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/version.rb:191:in `initialize': Malformed version number string 0.4.3-beta1 (ArgumentError)
	from /Library/Ruby/Gems/2.0.0/gems/docker-sync-0.4.3.pre.beta1/lib/docker-sync/upgrade_check.rb:32:in `new'
	from /Library/Ruby/Gems/2.0.0/gems/docker-sync-0.4.3.pre.beta1/lib/docker-sync/upgrade_check.rb:32:in `should_run'
	from /Library/Ruby/Gems/2.0.0/gems/docker-sync-0.4.3.pre.beta1/lib/docker-sync/upgrade_check.rb:14:in `run'
	from /Library/Ruby/Gems/2.0.0/gems/docker-sync-0.4.3.pre.beta1/tasks/sync/sync.thor:34:in `start'
	from /Library/Ruby/Gems/2.0.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
	from /Library/Ruby/Gems/2.0.0/gems/thor-0.19.4/lib/thor/invocation.rb:126:in `invoke_command'
	from /Library/Ruby/Gems/2.0.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
	from /Library/Ruby/Gems/2.0.0/gems/thor-0.19.4/lib/thor/base.rb:444:in `start'
	from /Library/Ruby/Gems/2.0.0/gems/docker-sync-0.4.3.pre.beta1/bin/docker-sync:14:in `<top (required)>'
	from /usr/local/bin/docker-sync:23:in `load'
	from /usr/local/bin/docker-sync:23:in `<main>'
```

Closing this PR 👍

Issue #370, opened 2017-05-08
Discuss if we should switch back to unison as default for now


EugenMayer:
Due to the increasing number of issues with native_osx like #366 things become complicated for users.

We probably should go back to unison as default until those things have stablized - i probably was to optimistic with native_osx - we need more time to get things stable.

Sure, native_osx stays available, but not as the default - what do you guys think?

michaelbaudino:
I totally agree we should switch back to unison as default 👍 

Also, what about making auto upgrade an "opt-in" option (users would have to use `--auto-upgrade` explicitely) ?

ignatiusreza:
hmmm.. ubuntu user here.. so, I can't really have a say.. :grin: .. though, I'll ask my fellow worker if they encountered any issue..

in term of auto upgrade being opt-in.. I guess we should consider it together with #304 by having an entry in the global config for auto upgrade behavior.. something like it's being `off`, `confirm`, or `on`..

rogamoore:
Good idea - let the user decide if he wants to try the new strategy.

Zakay:
I also suggest not requiring user input just displaying info/warning messages is enough. This to avoid breaking automated builds. Thing is even after you migrated to new configuration it still asks the questions, giving the feeling that the configuration that was just made still not up to date.

rwilliams:
My laptop battery will be sad but i'll be happy knowing that it works consistently if we switch back to unison.

EugenMayer:
@rogamoore the user can decide about his strategy since the start and even now, so what do you mean by that. We just decide about the default, the one you get if you did not decide yourself.

@Zakay that wont happen. The way docker-sync is upgraded, the way information is pushed so that users actually read the upgrade notes, read the changes, is required so the issue queue is not spammed with issues about the same thing over and over again. The point behind that is, that people not caring the RTFM will block other issues from getting done by flooding the queue, which will stall the project in the end - and i would like to avoid that and facing the consequences of broken updates, of issues and firefighting.

@rwilliams well we keep you posted, native_osx is by no means dead. But out of any reasons i expected it to be a smooth ride - the newer unexpected issues proven to be not talking the same language, so thats why i have to give myself defeated most probably, rolling back the default until the native_osx strategy is bullet proof ( or close to such )

codycraven:
[...]

PR #379, opened 2017-05-09
Don't enforce update by default

nporteschaikin:


EugenMayer:
please see https://github.com/EugenMayer/docker-sync/issues/370#issuecomment-300122946

Issue #319, opened 2017-04-27
RVM issue when upgrading


wulftone:
I tried to have it automatically upgrade and ran into this issue:

```bash
$ bundle exec docker-sync-stack start
     warning  The installation progress of docker-sync 0.3.0 has changed, brew is now mandatory - you need to uninstall unox ! :

_Please_ read :): https://github.com/EugenMayer/docker-sync/wiki/1.3-Upgrade-Guide

          ok  sudo rm -f /usr/local/bin/unison-fsmonitor && brew tap eugenmayer/dockersync && brew install eugenmayer/dockersync/unox
I will reinstall docker-sync for you using the above command (y/N) y
Password:
Your Ruby version is 2.0.0, but your Gemfile specified 2.2.2
Error: Another active Homebrew update process is already in progress.
Please wait for it to finish or terminate it to continue.
Your Ruby version is 2.0.0, but your Gemfile specified 2.2.2
/Users/Me/.rvm/gems/ruby-2.2.2/gems/bundler-1.10.5/lib/bundler/definition.rb:412:in `validate_ruby!': Your Ruby version is 2.0.0, but your Gemfile specified 2.2.2 (Bundler::RubyVersionMismatch)
	from /Users/Me/.rvm/gems/ruby-2.2.2/gems/bundler-1.10.5/lib/bundler.rb:123:in `setup'
	from /Users/Me/.rvm/gems/ruby-2.2.2/gems/bundler-1.10.5/lib/bundler/setup.rb:18:in `<top (required)>'
	from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
     warning  Could not find unison-fsmonitor (unox) binary in $PATH. Trying to install now
```

It appears to be wanting to use the system-level Ruby, not the one currently used by RVM.  I think it's ignoring the `bundle exec` somehow when running this command.

There is no other "active Homebrew update process" going.  I checked, it does this each time.  I ran the `sudo rm -f /usr/local/bin/unison-fsmonitor && brew tap eugenmayer/dockersync && brew install eugenmayer/dockersync/unox` but now the app just doesn't start... it gives the same Ruby version error.

EugenMayer:
[...]

PR #381, opened 2017-05-09
Fix RubyVersionMismatch error when using RVM

wulftone:
See https://github.com/EugenMayer/docker-sync/issues/319

### DIFF
--- a/lib/docker-sync/preconditions/preconditions_osx.rb
+++ b/lib/docker-sync/preconditions/preconditions_osx.rb
@@ -101,7 +101,12 @@ module DockerSync
 
       def unox_available
         if should_run_precondition?
-          Bundler.with_clean_env { `brew list unox` }
+          if defined? Bundler
+            Bundler.with_clean_env { `brew list unox` } 
+          else
+            `brew list unox`
+          end
+          
           unless $?.success?
             # unox installed, but not using brew, we do not allow that anymore
             if File.exist?('/usr/local/bin/unison-fsmonitor')
--- a/lib/docker-sync/preconditions/preconditions_osx.rb
+++ b/lib/docker-sync/preconditions/preconditions_osx.rb
@@ -101,7 +101,7 @@ module DockerSync
 
       def unox_available
         if should_run_precondition?
-          `brew list unox`
+          Bundler.with_clean_env { `brew list unox` }
           unless $?.success?
             # unox installed, but not using brew, we do not allow that anymore
             if File.exist?('/usr/local/bin/unison-fsmonitor')


EugenMayer:
I am hesitated to pull that one it because i do not understand it at all. Could you please explain?

wulftone:
Sure, from the [Bundler documentation](http://bundler.io/v1.3/man/bundle-exec.1.html#Shelling-out):
> Any Ruby code that opens a subshell (like system, backticks, or %x{}) will automatically use the current Bundler environment. If you need to shell out to a Ruby command that is not part of your current bundle, use the with_clean_env method with a block. Any subshells created inside the block will be given the environment present before Bundler was activated. For example, Homebrew commands run Ruby, but don't work inside a bundle:
```ruby
Bundler.with_clean_env do
  `brew install wget`
end
```

ignatiusreza:
I wonder, why do you still need to run the command via `bundle exec`? shouldn't rvm automatically activate the right ruby via `.ruby-version` & `.ruby-gemset`? I'm also using rvm, and I never have this issue.. I installed docker-sync into `@global` on the active ruby via `rvm @global do gem install docker-sync`.. and everything after that is just `docker-sync XXX` without needing `bundle exec`..

wulftone:
Yes that may be true but you are circumventing the benefits of using bundler (sandboxing) by doing that. We may have different projects using different versions of docker-sync, rather than forcing all to upgrade simultaneously

ignatiusreza:
isn't that's why rvm have gemset? by using different gemset on different project, you achieve the same level of sandboxing.. and if you don't want to install it under `@global` then it should still be okay to install it into the active gemset, but you'll need to remember to reinstall `docker-sync` everytime you do `bundle clean --force`..

wulftone:
I appreciate your words, but I don't really want to get into a bundler vs. gemset argument, there are a million of those on the internet already.  Anyway this project should support both in my opinion (like everyone else does).  My company's workflow is heavily dependent on bundler and we are not going to adopt gemsets ever.

ignatiusreza:
Sorry, I'm just trying to understand your situation and where you're coming from.. since this can open up a whole can of worm.. if we're to officially support rvm, then we need to make sure that it actually is working on every single spot where a version mismatch could happened.. sure, this PR fix one spot that we've stumbled upon for now, but who can guarantee that there're no other places that need similar treatment? and how will it affect maintenance, how can we make sure that it stays working?

wulftone:
Yes, I hear your concerns.  This exact scenario is called out in the bundler documentation because `brew` is a special case already.  Since you are choosing to integrate with `brew` I bet you already have version mismatches but gemsets handle it differently (if that is what you are all using!) and apparently without issue.  Other people may simply be using their system ruby (or system ruby in a container to have easier ruby version control).  Unless you expect `Bundler` to not be present for some reason (is there any Ruby project that does _not_ use bundler and would use `docker-sync`?) I am not sure if this even can cause any issue.  The bundler code is pretty simple to follow if you'd like to start tracing it from here https://github.com/bundler/bundler/blob/b6f5408e58b6ec6c24e004053ead6875d7a3f028/lib/bundler.rb#L286-L288

EugenMayer:
I am not using system ruby including the majority of users to not do either.

I am using rbenv, some use chruby and probably some use RVM, but a lot less use system ruby.

That said, i have never seen Bundler.with_clean_env in any code yet, i have seen https://makandracards.com/makandra/15649-how-to-discard-a-surrounding-bundler-environment explaining that this gets out of bundler for the execution.

brew is not ruby (bash) but might run ruby right - still that is not the duty of docker-sync to care about that right?

If you put your shell into the env of ruby > system or set that ruby to be your default globally, that is what homebrew will run in that env.

@wulftone so you understand why i do really block against this merge is, that this feels just not right at all. We have a couple of system shell calls and i have yet never seen a tool doing what you suggested to call anything on the shell. So i fear introducing new bugs doing that, thats why i am picky about that.

I really appreciatethat you explain the intention, but until know, i smell a local configuration issue, rather then a general RVM issue.

is there any way to clear that doubt?

wulftone:
I will have to think on it, thank you for your reply.

rwilliams:
With native_osx getting closer to stable in 0.4.5beta1 is this going to be an issue? I know it uses unison for the initial sync now but it no longer requires unox so the command referenced in the original bug report doesn't get run.

rwilliams:
Also just from poking around in brew source code maybe it's the use of `Kernel#exec` that's causing the issues?

wulftone:
You are right that it isn't an issue for `native_osx` but that appeared to have its own problems (as i see from some other issues here it isn't quite ready for prime-time).

I can detect if `Bundler` is defined if that would satisfy everyone, e.g.
```ruby
if defined? Bundler
  Bundler.with_clean_env { `brew list unox` } 
else
  `brew list unox`
end
```

Is unox going to be dropped completely once `native_osx` is shipped?

rwilliams:
My understanding is that unox is not at all required for native_osx as we aren't having unison listen to fsevents, we're only using unison for the initial sync.

rwilliams:
>Is unox going to be dropped completely once native_osx is shipped?

Are you asking if the unison strategy will still be available for OSX when `native_osx` becomes the default?

I don't really see any reason to use unison over `native_osx` on a mac is `native_osx` is behaving. `native_osx` is looking very close to being stable.

wulftone:
Yeah I was just curious if this would be an issue for anyone else who explicitly decided to not use `native_osx` strategy and go with `unison` instead even after `native_osx` is shipped for real.

EugenMayer:
Unison will not get removed, it's the freedom of choice

EugenMayer:
To be more precise, native_osx will become the default and is looking promising, still unison will be there, e.g. for cases you want to use docker-machine, you cannot use native_osx.

Docker-sync should be a solutions for all docker drivers, so people can chose whatever they like - so we will need to have a set of strategies - the best for the driver scenarios each.

d4m will have native_osx as the default - Hope that helps

wulftone:
Updated the PR. Is that better?

donnykurnia:
@wulftone if you are using docker, why you still using rvm/bundler on the OSX host? One of the reasons I use docker is to get rid of the messy rvm/rbenv and only have system ruby installed on my system. Ruby needed for my project are installed inside the docker image, not messing around the host. One of 'downside' is that I no longer executing `bin/rails c` but `docker-compose exec web bin/rails c` but I embrace it :-)

Having multiple `docker-sync` installed via bundler I think is not a good idea. You should only have 1 version installed and use it across projects. You will not need bundler to run `docker-sync` command if you install it in system ruby or rvm default ruby.

ignatiusreza:
I still have rvm on the host since not all of the projects that I want to get my hand on used docker.. especially when we're talking about open source projects..

though, agree that having docker-sync tied into bundler is not a good idea..it shouldn't even be in your `Gemfile`, since you need it on the host machine, while what you have in the `Gemfile` should go into the container..

EugenMayer:
I think thats fine for now. If we get negative reports we will remove it, but i think its fair enaugh

EugenMayer:

PR #389, opened 2017-05-12
Add a development console

michaelbaudino:
This adds a development console when running `bin/console` instead of having to type `/usr/bin/env ruby -r bundler/setup -r docker-sync -r pry -e Pry.start` because... you know... it's a lot handier ¯\\\_(ツ)\_/¯

The code is the same as the one generated by running `bundle gem <gem_name>` to create a new gem skeleton.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,6 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'rspec'
+group :test do
+  gem 'rspec'
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -14,13 +14,19 @@ GEM
   remote: https://rubygems.org/
   specs:
     backticks (1.0.0)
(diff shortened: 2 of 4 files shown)


EugenMayer:
I have no idea what that is but i am very interested in what you do with that? could you explain?

michaelbaudino:
Sure: when you're in the project root, you can simply start `bin/console` and it will give you a Ruby console with the code from the project already required. You can then manipulate your objects interactively.

Exemple session:
```
→ bin/console
[1] pry(main)> DockerSync::ProjectConfig.new(config_path: "../path/to/your/docker-sync.yml").unison_required?
=> false
```

michaelbaudino:
ℹ️ It's now automatically created by Bundler when generating a new gem skeleton using the `bundle gem` command.

EugenMayer:
That is super neat - wow! Learned a lot here.

PR #391, opened 2017-05-14
Refactor preconditions

### LINKED ISSUE #377: Refactoring Idea: Extract environment munging logic into its own namespace
Instead of having `#is_linux?`, `#is_osx?` etc under `Preconditions`, we should move it into its own namespace, e.g. `DockerSync::Environment`..

there're 2 options from there, either we have it included into classes where checking for environment is needed via `include DockerSync::Environment`, or we just call method function like `Environment.is_linux?`.. I lean toward the later approach since it feels cleaner, but it require more typing..

thoughts @EugenMayer & @michaelbaudino ?

PS: @michaelbaudino since you're working on #337, maybe you'll be interested to work on this too? if not I'll be doing it.. asking just to avoid potential double work.. :grin: 

PS2: btw, why does the method prefixed with `#is_`? I thought in ruby, [the standard](https://github.com/bbatsov/ruby-style-guide#bool-methods-prefix) is to just have `?` suffix, e.g. `Environment.linux?`..

EugenMayer:
@michaelbaudino did a huge job here, but i think we have to review it carefully.

That will be a huge step, thank you @michaelbaudino !

Closes #377


lib/docker-sync/dependencies.rb
```diff
@@ -0,0 +1,36 @@
...
+  File.join(File.dirname(__FILE__), 'dependencies', '**', '*.rb')
+].each { |f| require f }
+
+module DockerSync
+  module Dependencies
+    UNSUPPORTED_OPERATING_SYSTEM = 'Unsupported operating system. Are you sure you need DockerSync?'.freeze
+
+    def self.ensure_all!(config)
+      return ensure_all_for_mac!(config)   if Environment.mac?
+      return ensure_all_for_linux!(config) if Environment.linux?
```
EugenMayer: @rwilliams we do not need to cover WSL specifically here, do we?

lib/docker-sync/dependencies/docker_driver.rb
```diff
@@ -0,0 +1,20 @@
+module DockerSync
+  module Dependencies
+    module Docker
+      module Driver
+        def self.docker_for_mac?
+          return false unless Environment.mac?
+          return @docker_for_mac if defined? @docker_for_mac
+          @docker_for_mac = system('docker info | grep -q "Docker Root Dir: /var/lib/docker" && docker info | grep -q "Operating System: Alpine Linux"')
```
EugenMayer: actually, does ruby return @docker_for_mac here ?

lib/docker-sync/dependencies/fswatch.rb
```diff
@@ -0,0 +1,14 @@
+module DockerSync
+  module Dependencies
+    module Fswatch
+      def self.available?
+        return @available if defined? @available
```
EugenMayer: if not mac-OS we can return false, macOS only ```return false unless Environment.mac?```

lib/docker-sync/dependencies/package_manager.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      class << self
+        extend Forwardable
+        def_delegators :package_manager, :available?, :ensure!, :install_package
+      end
+
+      def self.package_manager
+        return @package_manager if defined? @package_manager
+        return @package_manager = PackageManager::Brew if Environment.mac?
+        raise(NO_SUPPORTED_PACKAGE_MANAGER)
```
EugenMayer: Actually that would not work under linux and it should? The point here is, we might only be able to actually pick ``apt-get``  here for linux @ignatiusreza what do you think. And if its not a debian based distro, we need to make it optional like "Cannot install package X", please do manually.

lib/docker-sync/dependencies/package_managers/brew.rb
```diff
@@ -0,0 +1,24 @@
+module DockerSync
```
EugenMayer: We should implement PackageManger::Apt - should be fairly easy at @ignatiusreza

lib/docker-sync/sync_process.rb
```diff
@@ -11,7 +11,7 @@
 require 'docker-sync/watch_strategy/unison'
 require 'docker-sync/watch_strategy/remotelogs'
 
-module Docker_Sync
+module DockerSync
```
EugenMayer: Thank you for doing that one :)

EugenMayer:
@michaelbaudino awesome job and @ignatiusreza nice suggestion with the ENV. Stuff is a lot cleaner now, huge improvement!

lib/docker-sync/dependencies.rb
rwilliams: Nope. It should behave just like linux.

lib/docker-sync/dependencies/docker_driver.rb
michaelbaudino: Yep: assignment expressions return the assigned value 👍

lib/docker-sync/dependencies/fswatch.rb
michaelbaudino: We can but that would make it dependant / coupled to `Environment`, which is something I'm trying to avoid. Do we really need it? (`find_executable0` is not an expensive call, if that's what you'd like to prevent)

lib/docker-sync/dependencies/package_manager.rb
michaelbaudino: That's an excellent point. Let me address this by adding `DockerSync::Dependencies::PackageManager::Apt` and `DockerSync::Dependencies::PackageManager::None` 👍

ignatiusreza:
Wow! This is huge! :+1: 

I'll take a look and give it a spin within today..

lib/docker-sync/dependencies/package_manager.rb
rwilliams: Is this requiring them to have brew installed even if they are using `native_osx` where it should not be needed?

lib/docker-sync/dependencies.rb
EugenMayer: great, consider done then

lib/docker-sync/dependencies/docker_driver.rb
EugenMayer: lesson learned..ruby sometimes scares me so i am extra sensitive. consider done

lib/docker-sync/dependencies/package_managers/brew.rb
michaelbaudino: done in last commits ✔️

lib/docker-sync/dependencies/fswatch.rb
EugenMayer: I see, no i do not mind it, but you already have that in https://github.com/EugenMayer/docker-sync/blob/47f1c6cf08f583a77355ce942a7a64f4c1841fd1/lib/docker-sync/dependencies/docker_driver.rb#L6 so i was not sure you want to keep it that dry

lib/docker-sync/dependencies/package_manager.rb
EugenMayer: I guess ::None is just this "you are not support, pleae install #{package} manuall" puts thing, right?

michaelbaudino: This is now addressed by last commits ✔️

lib/docker-sync/dependencies/fswatch.rb
michaelbaudino: Yep, it should not be in `Docker::Driver` neither imho, but it's the only way to avoid false positive of people **genuinely** using Docker on an Alpine Linux host 😕

lib/docker-sync/dependencies.rb
```diff
@@ -0,0 +1,37 @@
...
+      PackageManager.ensure!
+      Docker.ensure!
+
+      if config.unison_required?
+        Unison.ensure!
+      end
+
+      if config.rsync_required?
+        Rsync.ensure!
+        Fswatch.ensure!
```
ignatiusreza: technically `fswatch` is only needed when `watch_strategy` is `fswatch`.. I'm aware that this conditional is how it currently is in master, though, since we're refactoring.. should we actually split checking for `fswatch` into it's own branch? wdyt?

lib/docker-sync/dependencies/docker.rb
```diff
@@ -0,0 +1,23 @@
+module DockerSync
+  module Dependencies
+    module Docker
+      DOCKER_NOT_AVAILABLE = 'Could not find docker binary in path. Please install it, e.g. using "brew install docker" or install docker-for-mac'.freeze
```
ignatiusreza: this line is specific to osx.. should we refer user to `https://docs.docker.com/compose/install/` instead?

lib/docker-sync/dependencies/fswatch.rb
ignatiusreza: `fswatch` does exists on linux.. so we can't do `return false unless Environment.mac?`..

lib/docker-sync/dependencies/package_manager.rb
```diff
@@ -0,0 +1,18 @@
...
+  module Dependencies
+    module PackageManager
+      class << self
+        extend Forwardable
+        def_delegators :package_manager, :available?, :ensure!, :install_package
+      end
+
+      def self.package_manager
+        return @package_manager if defined? @package_manager
+        return @package_manager = PackageManager::Brew if PackageManager::Brew.available?
```
ignatiusreza: IIUC.. we want to make `brew` a hard requirement under osx.. so we need to update this to `return @package_manager = PackageManager::Brew if Environment.mac?`..

else, the next lines will continue be evaluated, and we will end up with `PackageManager::None`..

lib/docker-sync/dependencies/package_managers/none.rb
```diff
@@ -0,0 +1,23 @@
...
+        end
+
+        def self.ensure!
+          # noop
+        end
+
+        private
+
+        def install_cmd
+          raise(NO_PACKAGE_MANAGER_AVAILABLE)
```
ignatiusreza: should we tell which package were we trying to install? so that it's easier for user to realize which dependency are missing.. wdyt?

spec/lib/docker-sync/preconditions_spec.rb
```diff
@@ -1,142 +0,0 @@
-require 'docker-sync/preconditions/strategy'
```
ignatiusreza: shouldn't we rewrite this spec to follow the updated structure?

lib/docker-sync/dependencies.rb
michaelbaudino: Very good point: let's just keep `master` behaviour unchanged for this PR, but I'll do it in another PR 👍

lib/docker-sync/dependencies/docker.rb
michaelbaudino: Totally, let me change this 👍

lib/docker-sync/dependencies/package_manager.rb
michaelbaudino: Well, we **want** to end up with `PackageManager::None` actually: it will let the user use DockerSync in the case where all dependencies have been installed without brew (a user can have manually installed unison, fswatch, rsync or whatever dependency they need).

lib/docker-sync/dependencies/package_managers/none.rb
michaelbaudino: It's supposed to have been output the line before: https://github.com/EugenMayer/docker-sync/pull/391/files#diff-daa72dbe0caf7c551d89ad678f07c384R25

spec/lib/docker-sync/preconditions_spec.rb
michaelbaudino: We sure should! As seen with @EugenMayer, the specs will be in a separate PR 👍

lib/docker-sync/dependencies.rb
michaelbaudino: I actually implemented it ✔️

lib/docker-sync/dependencies/docker.rb
michaelbaudino: Done ✔️

lib/docker-sync/dependencies/package_manager.rb
ignatiusreza: Sure, sounds good

lib/docker-sync/dependencies/package_managers/none.rb
ignatiusreza: Ok, cool :+1:

michaelbaudino:
@ignatiusreza It's now on `master` if you want to play with it 👍
Thanks all for your reviews 🎩

PR #396, opened 2017-05-15
Fix license in gemspec to be consistent with https://spdx.org/licenses/

rwilliams:
Fix license in gemspec to be consistent with https://spdx.org/licenses/


michaelbaudino:
Very good point, thanks a lot @rwilliams  🎩

PR #403, opened 2017-05-20
set max_attempt option in `@config_syncs` using value from `@config_global`

donnykurnia:
This is a missing code to set the `max_attempt` value from `@config_global` into `@config_syncs`

Without this, sync process cannot read the global config value, thus will only using default value set in https://github.com/EugenMayer/docker-sync/blob/master/lib/docker-sync/sync_strategy/unison.rb#L185

### DIFF
--- a/lib/docker-sync/sync_manager.rb
+++ b/lib/docker-sync/sync_manager.rb
@@ -40,6 +40,13 @@ module DockerSync
           end
         end
 
+        # set the global max_attempt setting, if the sync-endpoint does not define a own one
+        unless config.key?('max_attempt')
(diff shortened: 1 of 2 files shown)


EugenMayer:
very good catch! thank you! 

@donnykurnia i invited you as a contributor since you are doing valuable work. Now matter you join the squad or not, thank you a lot for the time and work!

There are only a few rules:

- substantial work goes into a branch for PR - anything else can go onto the master directly
- be bold

And thats it

donnykurnia:
Thanks, @EugenMayer I appreciate and I'll try my best to help here.

PR #433, opened 2017-07-16
unison volume mount destination

pandanotbreathing:
probably there is a typo? or am I wrong?


EugenMayer:
Good catch, legacy code

Issue #439, opened 2017-07-18
Allow a prefix to be added to container names from command line


kducharm:
### Error/Feature Requestion/Docs
In trying to find a solution for #427, allowing for a prefix to be specified as a parameter to docker-sync would resolve the issue of using `docker-compose -p projectname` to have multiple sandboxes running on the same codebase by adding a similar parameter to docker-sync so that the volumes are uniquely named.

EugenMayer:
@kducharm this will need a load of work and also, i will not add this without having a maintainer for this. This is a little door to hell since we most probably cannot do a lot of assumptions and simplifications about names and things, that said, i am sorry, very sceptical about giving this my blessing right now.

kducharm:
Rather than trying to introduce this parameter, I was able to create a docker-sync.template.yml file and prior to running docker-sync, used a custom build script to generate the right container name using token replacement into a docker-sync.yml prior to starting everything.

PR #440, opened 2017-07-18
Added support for --sync-prefix to add to sync_name

kducharm:
A solution to #439 which when a --sync-prefix is added, volumes specified in the docker-sync.yml will be prefixed with the specified value, feedback appreciated.

`docker-sync start --sync_prefix=myprefix_`

### DIFF
--- a/lib/docker-sync/sync_manager.rb
+++ b/lib/docker-sync/sync_manager.rb
@@ -64,44 +64,45 @@ module DockerSync
       end
     end
 
-    def init_sync_processes(sync_name = nil)
+    def init_sync_processes(sync_name = nil, sync_prefix = '')
       return if @sync_processes.size != 0
+
       if sync_name.nil?
         @config_syncs.each { |name, sync_configuration|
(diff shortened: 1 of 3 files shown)


EugenMayer:
Will not merge this if we do not write tests for this one - this has been tried by @mickaelperrin  back in the days and the outcome was, AFAIR, its not possible. 

@mickaelperrin can you remember any details?

EugenMayer:
Any tests planned @kducharm ?

kducharm:
@EugenMayer I've been a bit slammed but definitely would like to see this added. I'm currently using this on 4 projects on the same machine simultaneously and sync has been working great. I'll see what integration tests descriptions I can come up with, I'm not a ruby developer so I might need some help.

kducharm:
@EugenMayer Rather than trying to introduce this parameter, I was able to create a docker-sync.template.yml file and prior to running docker-sync, used a custom build script to generate the right container name using token replacement into a docker-sync.yml prior to starting everything.

PR #459, opened 2017-09-08
Edit native_osx to support spaced directories.

### LINKED ISSUE #446: ~ resolving User directory with space in it breaks sync container startup
When using `~` to resolve the user path (useful when sharing environment with multiple users), if there is a space in the path, it breaks.
Assuming user's path is `/Users/name spaced/Documents`
Breaks
```
syncs:
  pks-web-sync:
    src: '~/Documents/'
```
Works
```
syncs:
  pks-web-sync:
    src: '/Users/name\ spaced/Documents/'
```

As an aside, it was rather random in how this came about. I was migrating one mac to another and I already had created a user of same name before starting Migration Assistant and it added ` 1` with the space to the username before creating the new account and then later deleted the second account which shared the name.

So I started with...
Mac 1:
`/Users/user`
Mac 2:
`/Users/user`

after Migration Assistant...
Mac 1:
`/Users/user`
Mac 2:
`/Users/user 1` - files and such all copied here

guille-moe:
Hello 🖐  !

This is my first PR, I don't really found a good way to test it, so any advice are very welcome !

Suppose to Fixes #446

### DIFF
--- a/lib/docker-sync/sync_strategy/native_osx.rb
+++ b/lib/docker-sync/sync_strategy/native_osx.rb
@@ -79,12 +79,12 @@ module DockerSync
             run_privileged = '--privileged' if @options.key?('max_inotify_watches') #TODO: replace by the minimum capabilities required
             say_status 'ok', 'Starting precopy', :white if @options['verbose']
             # we just run the precopy script and remove the container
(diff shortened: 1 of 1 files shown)


EugenMayer:
Your first PR looks pretty solid already, thank you. We might want to add a space-foldee integration test but more then that we are good to go here.

Issue #410, opened 2017-05-24
Not syncing both directions suddenly


nurtext:
### Error/Feature Requestion/Docs
All docker container are up, but not syncing between /host_sync and /app_sync.
/app_sync to /host_sync is working.

### Docker Driver
d4m

### Sync strategy
native_osx

### docker-compose.yml
```
version: '2'

services:
    api:
        container_name: api
        image: api
        build: .
        ports:
            - 8888:80
        volumes:
            - ./app:/srv/api/app
            - ./bin:/srv/api/bin
            - ./src:/srv/api/src
            - ./tests:/srv/api/tests
            - ./web:/srv/api/web
            - ./vendor:/srv/api/vendor
        environment:
            DATABASE_HOST: "127.0.0.1"
```

### docker-compose-dev.yml
```
version: '2'

services:
    api:
        volumes:
            - api-app-sync:/srv/api/app:nocopy
            - api-bin-sync:/srv/api/bin:nocopy
            - api-src-sync:/srv/api/src:nocopy
            - api-tests-sync:/srv/api/tests:nocopy
            - api-web-sync:/srv/api/web:nocopy
            - api-vendor-sync:/srv/api/vendor:nocopy

volumes:
    api-app-sync:
        external: true

    api-bin-sync:
        external: true

    api-src-sync:
        external: true

    api-tests-sync:
        external: true

    api-web-sync:
        external: true

    api-vendor-sync:
        external: true
```

### docker-sync.yml
```
version: '2'

options:
    verbose: true

syncs:
    api-app-sync:
        src: './app/'

    api-bin-sync:
        src: './bin/'

    api-src-sync:
        src: './src/'

    api-tests-sync:
        src: './tests/'

    api-web-sync:
        src: './web/'

    api-vendor-sync:
        src: './vendor/'
```

### `docker ps` output
```
CONTAINER ID        IMAGE                            COMMAND                  CREATED             STATUS              PORTS                            NAMES
72ee4eb5a02b        api                              "docker-php-entryp..."   2 minutes ago       Up 2 minutes        9000/tcp, 0.0.0.0:8888->80/tcp   api
7ec87ccaa84f        eugenmayer/unison:hostsync_0.2   "/entrypoint.sh su..."   15 minutes ago      Up 2 minutes        500/tcp                          api-vendor-sync
[...]

PR #465, opened 2017-09-25
Add native osx diagram

michaelbaudino:
This PR adds a [draw.io](https://draw.io) diagram to briefly and visually explain how the `native_osx` strategy works under the hood, as we realized it's not clear for everyone.

I did my best to make it clear, but I'm highly open (read: I'm almost begging 😄) for improvement suggestions and for correctness review 😬 

About embedding, I saved it as PNG+XML from the [Draw.io Chrome app](https://chrome.google.com/webstore/detail/drawio-desktop/pebppomjfocnoigkeepgbmcifnnlndla) which [seems to be](https://github.com/jgraph/drawio/wiki/Embed-Diagrams) the only way to embed it in markdown without having to export it to an image format after every change. Here is an exemple of embedding into GitHub Markdown (it should update automatically):

PS @EugenMayer I could not find a way to create diagonal color stripes as containers background as you suggested 😢


EugenMayer:
This is again such a huge effort and win by you @michaelbaudino ! i will for sure check the diagram and give you feedback as far as i can so we can make this part of the docs but also i would go further, sticking that to the front up docs to all people interested in docker-sync but are sceptical how it actually works and why its an improvement.

Awesome work!

michaelbaudino:
Thanks @EugenMayer 😊 

I just added a few notes related to configuration (in yellow-ish), which is useful by itself, but which also proves that changing the diagram and pushing it to the repo automatically updates the version embedded in Markdown ✨ 

Also, I added a Unison logo for fanciness.

michaelbaudino:
The final git log might get pretty dirty… we should probably squash commits before merging this branch, imho 💭

EugenMayer:
- i would cal the "docker mount" the "docker native linux mount" or similar, to indicate, no performance issues will arise
- i would add a pseudo path like /Users/me/myapp/src to "codebase" under osx to indicate thats a path on the local filesystem
- i would try to indicate that the host mount ( OSXFS ) is slow, using a snail or something - using something "fast" for the docker native linux mount"
- the docker mount to the app-conainter should have something noted like "native write and read speed"
- we need to make sure one can understand that host_sync <-> app_sync are synced and thus the write/read speed is decoupled from the OSXFS /host_sync ( slow ) and the linux native folder (app_sync/fast) .. i know that is hard and its even hard to explain, but thats the actual core of the whole idea right - thats what we really want to point out?

Thank you for your time!

EugenMayer:
@michaelbaudino what do you think about the proposed changes, any comments?

EugenMayer:
hehe, all the icons, awesome. Now the last thing i would have is the unison "arrow".. all the other arrows are the same, but are direct mounts, this is a sync. And thats the really key point in our design here.

So how do we make it visual, that this is "synced" not "mounted". We could use this refresh icon https://www.iconfinder.com/icons/186384/refresh_sync_icon and put it beween, so arrow out, into the refresh icon, and then arrow in/out into app-sync

host_sync <-> refresh <-> app_sync

what do you think?

michaelbaudino:
Exactly, changing the Unison arrow to make it clear it's a sync is what I'm trying to achieve at the moment 👍 

Also, I'd like to emphasize the fact that development applications in the app container (in my (kn)own use case `rails server` or `webpack` mostly but I guess any other application server will do the same) will perform a shitload of read operations, and that those read operations are **not** going through OSXFS (thanks to Unison that only syncs changes) because I feel that it's one of the main reasons this setup is powerful. First question: am I right? Second question: should we put it on this diagram? Third question: any idea how? 😄

EugenMayer:
Yes you are right, ruby(ror)/php frameworks have the same DNA, a lot of file, not compiled, runtime parsed. Means that during every rake / web request, the whole code is parsed (read) and then executed ( in the memory, so no impact ).

This is why you can improve PHP frameworks by using opecache with "save the parser output" kind of setup, so this reads are not happening. Just this means, that syncing a file will not be reflected in the app, since its not re-parsed. You have to clear the opcache cache MANUALLY to have this reflected. This is PITA, but also, that the first bootstrap after that flush is ultra slow, since now everything is reread.

The point that even you do not understand this fully or at least have question is the best sign that we REALLY need to explain that. 

That is also the reason, why the sync-params of d4m are not helping, because they do not optimize reads and stuff.

So your whole post has nailed it - what you said - basically :)

michaelbaudino:
I'm quite happy with the current version, what do you think? (sorry the sync icon you proposed was circle-ish, and thus not easy to integrate)

I'm thinking of doing another similar diagram but *without* DockerSync to show this "shitload of read operations" issue.

michaelbaudino:
I may almost add `1`, `2` and `3` labels on the yellow notes, as basic usage steps 🤔

EugenMayer:
I would really like to replace that yellow unison sign with the real "sync" icon which is known for doing exactly that. What do you think?

michaelbaudino:
Like that? I must admit I'm not a big fan, but if you like it, let's keep it 👍

EugenMayer:
@michaelbaudino usually i would just say, revert it back since you are simply the author of this and i would not bother fighting about graphics at all.

But..this really makes it clear. I am very open to take some more opinions in on this, but i really like that the core point is now visible.

If you like, i would even add something like "sync decoupled read/write speed from OSXFS".

In any way @michaelbaudino this is such a huge milestone you are adding to docker sync...

guille-moe:
@EugenMayer I'm more fan of the Bidirectionnal arrows icon with the Unison `U` than the sync icon, anyway, @michaelbaudino your diagram is great and clear !

michaelbaudino:
Another try, with a new layout, hopefully clearer. Let me know what you think 😊

EugenMayer:
i lake that one too. We should add the word "sync" to the unison part - do not forget, not everybody or even nobody knows what unison does or is - so its not self explainatory.

Thats why i also dislike the logo only - its just for us, we know it. But the diagram is more for anybody else - thats why i would love to pick common meaning icons / pictures ;/

michaelbaudino:
Gotcha 👍

EugenMayer:
Looks great. I will try to align stuff a bit later but its really good.

michaelbaudino:
Thanks 🤗

FYI, I tried to keep the yellow notes **not** aligned on purpose (to emphasize that they're some kind of overlay of the diagram 😏).

mike-es:
This is a great diagram. It really helped my understand how the sync works.  It might be useful to add FSEvents/inotify given #410.

EugenMayer:
@michaelbaudino all credits to you, especially if you consider what @mike-es did with that knowledge on #410 - you noth are heros!

EugenMayer:
@michaelbaudino i would actually leave the diagram as is and merge that one in. We can still improve it over time, when people express their ideas and concerns about what they see. We might have a more neutral POV then and get a better general result. What do you think?

If you are happy with that, just press the merge button, its all yours!

michaelbaudino:
Merged 🎉 Feel free to link this image from any wiki page, @EugenMayer 👍 The link in Markdown is:

```markdown
![DockerSync native_osx strategy overview](https://raw.githubusercontent.com/EugenMayer/docker-sync/master/doc/native_osx.png)
```

EugenMayer:
Oh there will be a looooot of links to that little diagram, oh boy it will be extensive :)

Wonderful work! @michaelbaudino once again, thank you for this milestone

PR #474, opened 2017-10-20
Updating the stack start note

FlorianDr:
Updating the note, when starting the stack. `docker-sync --daemon` gives me `Could not find command "__daemon"`


EugenMayer:
Good catch, thank you sir!

PR #475, opened 2017-10-23
Fixing prefer dest

jerome-arzel:
Prefer parameter should contains the same destination as the unison command. #476

### DIFF
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -135,7 +135,7 @@ module DockerSync
         case @options.fetch('sync_prefer', 'default')
           when 'default' then "-prefer '#{@options['src']}' -copyonconflict" # thats our default, if nothing is set
           when 'src' then "-prefer '#{@options['src']}'"
-          when 'dest' then "-prefer '#{@options['dest']}'"
+          when 'dest' then "-prefer 'socket://#{@options['sync_host_ip']}:#{sync_host_port}'"
           else "-prefer '#{@options['sync_prefer']}'"
         end
       end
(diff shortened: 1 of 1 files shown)


EugenMayer:
sorry way to half baked for any inclusion. This is a "works only for me patch".

jerome-arzel:
Can you reconsider this pull request and guide me to improves it? I'm not sure why it was closed.

EugenMayer:
See the comments above and try to understand them or explain the changes so i can understand them :)

jerome-arzel:
Sure, there was more details in issue #476 

I reexplain all from the beginning, with this docker-file.yml:

~~~
version: '2'
options:
  verbose: false
syncs:
  project-sync:
    src: '../project'
    sync_strategy: 'unison'
    sync_prefer: 'dest'
    watch_strategy: 'dummy'
~~~

When I start docker-sync, I have the following error from Unison:

~~~
     success  Unison server started
     command  docker inspect --format='{{(index (index .NetworkSettings.Ports "5000/tcp") 0).HostPort}}' project-sync
     command  unison '/Users/xxx/Projects/project' -auto -batch -prefer '/app_sync' socket://127.0.0.1:32769
       error  Error starting sync, exit code 0
     message  
     message  Fatal error: /app_sync (given as argument to 'prefer' or 'force' preference)
is not one of the current roots:
  /Users/xxx/Projects/project
  socket://127.0.0.1:32769
     success  Starting Docker-Sync in the background
~~~

That's because Unison want to see the source or the target path on the `-prefer` parameter. Currently in docker-sync this works with sync_prefer: source, but not with sync_prefer: target, because we pass the path from the configuration, and not the real path used when we call unison (socket://127.0.0.1:32769). See? no big deal, just a simple bug :)

EugenMayer:
thank you for you patience and the brief explanation, i was completely missreading the diff - sorry.

PR #478, opened 2017-10-26
Fix D4M edge strategy detection

tripox:
With the newest version of Docker for Mac edge (17.10.0-ce-mac36), `docker info | grep 'Operating System'` outputs: `Operating System: Docker for Mac`.

This results in docker-sync choosing the 'unison' strategy which seems broken.


tripox:
This should be against the 0.4.6 tag but I'm not able to select it.

EugenMayer:
could you rebase the branch so we do not have merge conflicts?

EugenMayer:
great contribution ,thank you!

PR #487, opened 2017-11-06
Use pgrep for Docker for Mac detection

tripox:
A way more simple, cleaner and reliable way to detect if the process `com.docker.osx.hyperkit.linux` is running.

### DIFF
--- a/lib/docker-sync/dependencies/docker_driver.rb
+++ b/lib/docker-sync/dependencies/docker_driver.rb
@@ -5,7 +5,7 @@ module DockerSync
         def self.docker_for_mac?
           return false unless Environment.mac?
           return @docker_for_mac if defined? @docker_for_mac
-          @docker_for_mac = system('ps x | grep MacOS | grep -q com.docker.osx.hyperkit.linux')
+          @docker_for_mac = system('pgrep -q com.docker.osx.hyperkit.linux')
         end
 
         def self.docker_toolbox?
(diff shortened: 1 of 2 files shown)


EugenMayer:
@tripox is thes yet a nother way to detect if its d4m or not, right? I would delay this one after 0.5.0 because the current one works and i would not really like to get this into a new cycle again. What is your though about that?

tripox:
@EugenMayer, that's fine with me. :) 
The difference is that with ps, we have to grep for `MacOS` to avoid getting the `grep` process itself (because then it would always return exit code 0). And _IF_ they change the folder name, then this will break. With `pgrep` we can avoid that. :)

michaelbaudino:
This looks great. Is `pgrep` available in the system by default?

jerome-arzel:
It seems yes, pgrep is /usr/bin/pgrep

tripox:
@EugenMayer would it make sense to throw it in a minor release?

EugenMayer:
@tripox i would say no from my POV, since we have no bug to fix right now? The current method works, so i would do that in 0.6.0 where we can argue about a "upgrade". What do you think?

tripox:
That makes sense. I agree, @EugenMayer. :)

EugenMayer:
A soon as this becomes rather a bug, we will make it part of 0.5.x for sure

EugenMayer:
On my ENV, Sierra with edge, it would be `pgrep -q com.docker.hyperkit` and not `pgrep -q com.docker.osx.hyperkit.linux` .. can anybody confirm this?

tripox:
Yup, that apparently changed again. com.docker.hyperkit should stay though. Both was there before.

EugenMayer:
using `pgrep -q com.docker.hyperkit` for now for the bet thank for the patience

PR #495, opened 2017-11-09
Update readme with d4m issue

MichaelHindley:
As a developer working on osx, I would default to the native_osx strategy and be unaware of the issues it has, leading to potentially hours of troubleshooting.

This pr suggests to point osx developers towards a currently stable strategy instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,9 @@
     Thank you for all the feedback and support i already received!
     Docker-sync has been improved by all of you in huge ways!
 
+NOTE: [Due to bind mount inotify events not delivered after container restart](https://github.com/docker/for-mac/issues/681) the native_osx strategy on macOS may not work as expected.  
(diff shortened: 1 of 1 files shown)


EugenMayer:
You claim is false, your assumption is false and the solution is false. In addition, that we will not put anything of those in the readme, as you see, we do not put the latest "howtos" on there. Thats the wiki or the issue queue.

Saving time is good, but what you do is - you think your issue happens to anybody right away and we need a warning. No its now

MichaelHindley:
Tried this on 3 laptops running high sierra now, issue is present on all and is resolved in the exact same way- using rsync strategy instead of native_osx. I don't see how the claim is false, it's explicitly pointed out by others in #410 as well, as is the solution to use rsync.

If you'd rather people that try docker-sync with default settings just give up and feel let down, as almost happened to me, then thats your prerogative I guess.

Either way, try looking at things from a users perspective instead of a technical one, once in a while.

Leaving this toxic conversation now.

PR #496, opened 2017-11-10
Fix Docker configuration detection in RSpec

michaelbaudino:
That's a small fix to correctly detect if Docker For Mac has the temporary directory shared with HyperKit when running the specs.

It was previously failing in case you had `/var` being a symlink to `/private/var` (which was the case on my system, no idea why 🤷‍♂️). Adding `File.realpath` here fixes it by checking the *actual* path.

Not a big deal, no rush: it only determines whether to print a warning message or not before the test suite (_i.e._ merging this or not won't change the test suite result).


EugenMayer:
Very cool, nice we fixed this now.

PR #521, opened 2018-01-22
Use Monit to auto restart if CPU usage high in Native OSX strategy

### LINKED ISSUE #497: High CPU usage and syncing stops with native OSX when files are being created and removed
### Error

- When creating and removing a file such as when checking out a branch in git or other scenarios, the sync container will start using very high CPU usage and syncing will stop happening
- CPU usage stays at 100%, and memory usage also is very high
- The only way is to restart the sync container or the unison process inside the container
- As far as I can tell, there seems to be a bug in `unison` which causes this issue, but I'm not sure if docker-sync or docker is related. I am reporting this issue to see if there are others that can help with this as I continue investigating as well
- I have created a repro which should be able to reproduce this issue. It reproduces on my machine pretty reliably: https://github.com/ricmatsui/docker-sync-cpu-repro
- Steps to use the repro
  - Install gems
  - Run `./repro.sh`
  - It will run a series of tests with a larger and larger test file to get the issue to happen
  - It creates a file, waits 1 second, then removes the file and waits 1 second, and does this 15 times
  - Then it checks the CPU usage 5 times and reports if it sees high CPU usage
  - If it does not see high CPU usage, it increases the size of the file 10x and repeats until it reaches a 100M length file
  - The issue reproduces regularly with 100K lines in the file
  - Let me know if this does not reproduce the issue

Recording: Repro at 2:20

[asciicast](https://asciinema.org/a/147070)

### Docker Driver

d4m

Version 17.09.0-ce-mac35 (19611)
Channel: stable
a98b7c1b7c

### Sync strategy

native_osx
docker-sync (0.4.6)
eugenmayer/unison:hostsync_0.2

### your docker-sync.yml

https://github.com/ricmatsui/docker-sync-cpu-repro/blob/master/docker-sync.yml
```
version: '2'

syncs:
  reprocpu_sync:
    image: 'eugenmayer/unison:hostsync_0.2'
    src: '.'
    sync_strategy: native_osx
    sync_prefer: src
```

### OS

OSX 10.12.6

MacBook Pro 15-inch Mid 2015
2.5 GHz Intel Core i7
16 GB 1600 MHz DDR3

ricmatsui:
- Fixes #497
- Issue has repro steps
- Adding monit to Native OSX image done here: https://github.com/EugenMayer/docker-image-unison/pull/11
- This PR adds ability to control three configuration options from the config file

---

- `monit_enable` - Control if `monit` is used, default `false`
- `monit_interval` - How often in seconds `monit` performs check, default `5`
- `monit_high_cpu_cycles` - How many cycles of `monit` checks need to have high cpu usage (>50%) to trigger a restart, default `2`

### DIFF
--- a/example/docker-sync.yml
+++ b/example/docker-sync.yml
@@ -117,9 +117,9 @@ syncs:
     # optional: default is fswatch, if set to disable, no watcher will be used and you would need to start the sync manually
     watch_strategy: 'fswatch'
 
-    # monit is used to monitor the health of unison in the native_osx strategy and can restart unison if it detects a problem
-    # optional: use this to switch monit monitoring off
-    monit_enabled: true
+    # monit can be used to monitor the health of unison in the native_osx strategy and can restart unison if it detects a problem
+    # optional: use this to switch monit monitoring on
+    monit_enabled: false
     # optional: use this to change how many seconds between each monit check (cycle)
     monit_interval: 5
     # optional: use this to change how many consecutive times high cpu usage must be observed before unison is restarted
--- a/example/docker-sync.yml
+++ b/example/docker-sync.yml
@@ -116,3 +116,11 @@ syncs:
     watch_args: '-v'
     # optional: default is fswatch, if set to disable, no watcher will be used and you would need to start the sync manually
     watch_strategy: 'fswatch'
+
+    # monit is used to monitor the health of unison in the native_osx strategy and can restart unison if it detects a problem
+    # optional: use this to switch monit monitoring off
+    monit_enabled: true
+    # optional: use this to change how many seconds between each monit check (cycle)
+    monit_interval: 5
+    # optional: use this to change how many consecutive times high cpu usage must be observed before unison is restarted
+    monit_high_cpu_cycles: 2
--- a/lib/docker-sync/sync_strategy/native_osx.rb
+++ b/lib/docker-sync/sync_strategy/native_osx.rb
@@ -65,6 +65,8 @@ module DockerSync
           env['OWNER_UID'] = @options['sync_userid'] if @options.key?('sync_userid')
         end
 
+        env['ENABLE_MONIT'] = @options.fetch 'enable_monit', true
+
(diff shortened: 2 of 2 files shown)


EugenMayer:
@ricmatsui great contribution, thats really awesome! Could you please update the docs also, that means

https://github.com/EugenMayer/docker-sync/wiki/2.-Configuration
and in parallel https://github.com/EugenMayer/docker-sync/blob/master/example/docker-sync.yml

It would be great to have one monit example in the boilerplate, what do you think? https://github.com/EugenMayer/docker-sync-boilerplate 

I also would like to have a advanced topic on https://github.com/EugenMayer/docker-sync/wiki about that - i created https://github.com/EugenMayer/docker-sync/wiki/native_OSX-advanced---monitoring

Could you write some lines about why this is useful, what it does and link to the configuration? That would be awesome.

ricmatsui:
@EugenMayer Thanks, I have added the docs now in those locations and here's an example, too: https://github.com/EugenMayer/docker-sync-boilerplate/pull/11

EugenMayer:
great, that's really well done. last topic to discuss is the default, right now it's on. I would suggest not doing that for now, or we would be forced to release a new minor... that would delay that feature. in addition we do not have a wide field test yet and we do not really now which cycles are a sane 'mass compatibile' value, with possible tradeoffs on reaction time but for all (anybody can adjust those for quicker reaction and faster cleanup). I pledge for not being on by default for the 0.5.x series, but becoming on by default with 0.6

what are your thoughts on this gentleman?

ricmatsui:
@EugenMayer Yeah that makes sense, it'll be good to get wider testing sooner. I have switched to off by default in the PRs and docs now!

EugenMayer:
that's great, this is a long awaited addition to docker-sync, thank you for working your way through that!

I will release this in the next few days with 0.5.3 - any concerns?

ricmatsuitile:
@EugenMayer No problem, thanks for pushing this out! Sounds great.

PR #535, opened 2018-03-07
Foreground typo?

clementblanco:


EugenMayer:
Hehe, good catch - it changed to be on by default..which was the cause of that switch being inverse. Thanks!

PR #543, opened 2018-03-14
Fix monit_enabled typo / docs mismatch

nilbus:
PR #521 had a typo that causes the `monit_enabled` option to have no effect.

As a forward-compatible temporary workaround, one can set both options:

```yaml
monit_enable: true
monit_enabled: true
```

/cc @ricmatsui

### DIFF
--- a/lib/docker-sync/sync_strategy/native_osx.rb
+++ b/lib/docker-sync/sync_strategy/native_osx.rb
@@ -66,7 +66,7 @@ module DockerSync
         end
 
         monit_options = {
-          monit_enable: 'MONIT_ENABLE',
+          monit_enabled: 'MONIT_ENABLE',
           monit_interval: 'MONIT_INTERVAL',
           monit_high_cpu_cycles: 'MONIT_HIGH_CPU_CYCLES',
         }


EugenMayer:
actually the variable should be named monit_enable? https://github.com/EugenMayer/docker-image-unison/blob/feature/unison-hostsync/entrypoint.sh#L72 .. which also is reflexted here https://github.com/EugenMayer/docker-sync/blob/master/lib/docker-sync/sync_strategy/native_osx.rb#L69 so i would suggest `monit_enable` is just right?

EugenMayer:
Nevertheless i think the docs are wrong https://github.com/EugenMayer/docker-sync/wiki/2.-Configuration in here its `monit_enabled:` and should rather be `monit_enable` @ricmatsui could you comment on your intention here? thanks

ricmatsuitile:
Thanks @nilbus sorry for the confusion, @EugenMayer the intention was `monit_enable`, so it's mostly the docs and the boilerplate that are out of sync, let me know if I can help with this.

ricmatsui:
Thanks @nilbus sorry for the confusion, @EugenMayer the intention was `monit_enable`, so it's mostly the docs and the boilerplate that are out of sync, let me know if I can help with this.

nilbus:
You're right, this PR does introduce a mismatch between the `monit_enabled` config option and the `MONIT_ENABLE` env var [in docker-image-unison][env].

[env]: https://github.com/EugenMayer/docker-image-unison/blob/e4e4fdf8b7c6554831dae9c04a5936164f5d0d78/entrypoint.sh#L72

I think I was injecting my own opinion here that it ought to be monit_enabled everywhere, mostly because _enabled_ is much more consistent with option names in other libraries (e.g. apache `sites_enabled`, browser DOM `document.fullscreenEnabled`, and lots of other examples. As a boolean config option, I would expect the name to either be an adjective (`monit_enabled`), or a verb with an object (`enable_monit`). `monit_enable` feels like bad English grammar to me.

That said, perhaps there's value in consistency in the name between this repo's config and the env var in docker-image-unison. If you think it should be `monit_enable`, I'll go ahead and update the PR. Let me know :)

EugenMayer:
i`ll just stay with the intend of the author.

Fixed docs, fixed example/docker-sync.yml and the boilerplate. Still thanks a lot nilbus for point that out!

PR #563, opened 2018-03-28
Change image modification example

rodrigoaguilera:


EugenMayer:
hello @rodrigoaguilera thanks for digging that one up. I am sorry to not merge this since this is just a bug, no need to document it. I will release a new version with a fixed image string to fix that rather then documenting how to set the image manually.

The point with doing so is, that next time we update an image, this line will force you to use the old one thus force you to maybe have issues with docker-sync. Hope you understand

PR #577, opened 2018-05-14
replaced &> /dev/null with 2>&1 > /dev/null

Powerhamster:
When running with WSL default shell is `sh` and not `bash`, so output is not captured and printed to STDOUT. The operator `&>` does only work with `bash`


EugenMayer:
thank you a lot, that's not even known to me that dash is not supportong that. thanks for the pr!

Issue #356, opened 2017-05-06
docker-sync stop throwing error after update to 0.4.2


tanc:
### Error/Feature Requestion/Docs
`docker-sync stop` throwing error after update to 0.4.2

```
$ docker-sync stop
        ok  Stopping sync container php-oxfam-sync
/Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/lib/docker-sync/watch_strategy/unison.rb:30:in `kill': no implicit conversion from nil to integer (TypeError)
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/lib/docker-sync/watch_strategy/unison.rb:30:in `stop'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/lib/docker-sync/sync_process.rb:93:in `stop'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/lib/docker-sync/sync_manager.rb:135:in `block in stop'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/lib/docker-sync/sync_manager.rb:134:in `each'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/lib/docker-sync/sync_manager.rb:134:in `stop'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/tasks/sync/sync.thor:65:in `stop'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/thor-0.19.4/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/thor-0.19.4/lib/thor/base.rb:444:in `start'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/gems/docker-sync-0.4.2/bin/docker-sync:14:in `<top (required)>'
	from /Users/tanc/.rvm/gems/ruby-2.4.0/bin/docker-sync:22:in `load'
[...]

PR #578, opened 2018-05-27
Fix exception thrown on docker-sync stop with unison setup

asentner:
This will address the `no implicit conversion from nil to integer (TypeError)` exception that is thrown when using docker-sync stop as mentioned in #356.

I do not have a ton of experience with Ruby but it appeared the object was nil and not what was expected by the Process module. The unison process still start and stops normally after the fix.

Hope to see this make it in to the next release!

### DIFF
--- a/lib/docker-sync/watch_strategy/unison.rb
+++ b/lib/docker-sync/watch_strategy/unison.rb
@@ -27,8 +27,11 @@ module DockerSync
       end
 
       def stop
-        Process.kill 'TERM', @watch_fork
-        Process.wait @watch_fork
+        # Make sure @watch_fork is not nil otherwise a TypeError is thrown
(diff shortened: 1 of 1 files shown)


EugenMayer:
thank you @asentner!

immoreel:
I'm still experiencing this behaviour :-(

asentner:
You running from master? I don't think this made it into a release yet.

arosenhagen:
so why not creating an up-to-date release?!

edvinasnet:
can we add this fix to release?

asentner:
Looks like it made it to release. Updated to 0.5.9 today and the errors are gone now and that it was merged into 0.5.8 👍 

Thanks!

Issue #588, opened 2018-07-17
[idea/discussion] Support 1-way sync from container to host


netiul:
In our project we have some large folders that originate from the container. Changes in these folders are only made in the container and are very useful to have on the host machine. F.e. dependency folders like `vendor` and `node_modules`, or a `cache` folder.
Installing and updating dependencies we perform on the container and not on the host machine (less dependencies required on the host machine, less confusion on where to perform commands). 
Including these folders in the default synchronization process is a bad idea, but are very useful to have on the host for f.e. code completion and inspection.

Docker by default offers mounting volumes readonly by the container, but that's exactly the other way around.
The `rsync` strategy also syncs from host to container and that makes sense when writing code.

What about adding a reverse `rsync` strategy that actually syncs back changes from the container to the host, ignoring changes made on the host? It would lessen the long initial sync when using `native_osx`/`unison` strategy significantly. Any thoughts on this?

Edit: I just saw that @talya brought up the same idea https://github.com/EugenMayer/docker-sync/issues/585#issuecomment-404467257

EugenMayer:
I understand what you are looking for and it could be useful to some.

for my POV I want be able to put any work into that, but if you do, I am happy to review any pull requests.

We tend to have a rule of thumb here, if an idea does not get any traction after one month, we close it until someone picks it up... it gets reopened. I would like to apply this here too.

looking forward to reviewing your stuf

talya:
@netiul - awesome to hear you have a similar use case :)

Have you tried this approach manually yet?
I have currently switched my efforts to poc a similar setup but with VirtualBox, but I hope to be able in the future to tryout this rsync from inside the container and see if it's viable.

Do update if you have any success or other useful findings.

netiul:
@EugenMayer @talya Glad to see some thumbs up. I can work on this, but not before the next 4 weeks (will have some vacation in between).

barat:
I just want to say, that this will be great. I'm doing this kid of sync via Makefile now (and I have filewatch in phpstorm to execute it on save to have services registered by annotations discovered)
```
refresh-cache:
	@curl -k --output /dev/null --silent --head --fail https://project.local/

sync-cache:|refresh-cache
	@mkdir -p ./cms/app/cache/dev
	@$(eval CONTAINER:=`docker-compose ps -q service-name`)
	@docker cp $(CONTAINER):/app/cms/app/cache/dev/profiler/. ./cms/app/cache/dev/profiler/
[...]

PR #602, opened 2018-08-30
WIP: Provide the means to support 1-way synchronization of files from cont…

netiul:
**TL;DR:**
Provides the means to support 1-way synchronization of files from container to host
 
See https://github.com/EugenMayer/docker-sync/issues/588
 - Starts an rsync daemon on the host listening on the specified port
 - Starts a container that synchorizes the volume upon fs change events using the rsync command targeting the started rsync daemon on the host

This gem includes the current changes: [docker-sync-dev.gem.zip](https://github.com/EugenMayer/docker-sync/files/2336699/docker-sync-dev.gem.zip)

**Status**
Successfully using this strategy for a week now with 
 - a sync-point with 13560 files (96M)
 - a sync-point with 16797 files (411M)
 - a sync-point with 37465 files (319M)

Startup is not slower and files are synced within seconds.

**1-way sync**
The idea is to provide a way to synchronize certain directories from **container to host**. Exactly the opposite of what the rsync-strategy does (from host to container). This can be useful for directories that are generated/changed **only** in the container.

When the `reverse_rsync`-strategy is configured, basically this is done:

- an rsync daemon is started on the host listening on the specified port by `sync_host_port`, it's configured with the docker-sync.yml `src` value as a [module](https://download.samba.org/pub/rsync/rsyncd.conf.html).
- a [zluiten/fswatch-rsync](https://hub.docker.com/r/zluiten/fswatch-rsync/) container is started with the configured volume mounted. This container watches file changes by `fswatch` and starts an rsync command upon an fs event.

**!!WARNING!!** This **will** delete files on the host that are not inside the container as this is a 1-way sync from container to host and assumes that the container is always leading.

Please note that an rsync daemon is started for each configured `reverse_rsync` container, therefore each `sync_host_port` must be unique. (This could possibly be optimized by reusing an already started rsync daemon and just add a new module, but it has some side effects.)

**docker-sync.yml**
```yaml
version: "2"

options:
  verbose: true
  max-attempt: 500
syncs:
  #IMPORTANT: ensure this name is unique and does not match your other application container name
  project-sync: # tip: add -sync and you keep consistent names as a convention
    src: ./repositories/project
    sync_strategy: 'native_osx'
    sync_userid: '33'
    sync_excludes:
      - 'vendor'
  vendor-sync: # tip: add -sync and you keep consistent names as a convention
    src: ./repositories/project/vendor
    sync_strategy: 'reverse_rsync'
    sync_userid: '33'
    sync_host_port: 20873
    sync_excludes:
      - '/some-uninteresting-vendor'
```

**TODO**
- [ ] Add unit tests
- [ ] Validate configuration upon `sync:start` (f.e. detect in use `sync_host_port` values)
- [ ] Review rsyncd pid file location?
- [ ] Add some useful code comments
- [ ] Update documentation
- [ ] Add some performance test results

### DIFF
--- a/lib/docker-sync/config/project_config.rb
+++ b/lib/docker-sync/config/project_config.rb
@@ -57,7 +57,7 @@ module DockerSync
     def rsync_required?
       # noinspection RubyUnusedLocalVariable
       config['syncs'].any? { |name, sync_config|
-        sync_config['sync_strategy'] == 'rsync'
+        sync_config['sync_strategy'] == 'rsync' || sync_config['sync_strategy'] == 'reverse_rsync'
       }
     end
 
@@ -151,6 +151,7 @@ module DockerSync
       def default_watch_strategy(sync_config)
         case sync_strategy_for(sync_config)
         when 'rsync' then 'fswatch'
+        when 'reverse_rsync' then 'dummy'
         when 'unison' then 'unison'
         when 'native' then 'dummy'
         when 'native_osx' then 'remotelogs'
--- a/lib/docker-sync/sync_process.rb
+++ b/lib/docker-sync/sync_process.rb
@@ -1,6 +1,7 @@
 require 'thor/shell'
 # noinspection RubyResolve
 require 'docker-sync/sync_strategy/rsync'
+require 'docker-sync/sync_strategy/reverse_rsync'
 require 'docker-sync/sync_strategy/unison'
 require 'docker-sync/sync_strategy/native'
 require 'docker-sync/sync_strategy/native_osx'
@@ -44,6 +45,8 @@ module DockerSync
       case @options['sync_strategy']
       when 'rsync'
         @sync_strategy = DockerSync::SyncStrategy::Rsync.new(@sync_name, @options)
+      when 'reverse_rsync'
+        @sync_strategy = DockerSync::SyncStrategy::ReverseRsync.new(@sync_name, @options)
       when 'unison'
         @sync_strategy = DockerSync::SyncStrategy::Unison.new(@sync_name, @options)
       when 'native'
--- a/lib/docker-sync/sync_strategy/reverse_rsync.rb
+++ b/lib/docker-sync/sync_strategy/reverse_rsync.rb
@@ -1,0 +1,192 @@
+require 'thor/shell'
+require 'terminal-notifier'
+require 'tmpdir'
+require 'timeout'
+
+module DockerSync
+  module SyncStrategy
+    class ReverseRsync
+      include Thor::Shell
+
+      @options
+      @sync_name
+      @watch_thread
+      @rsyncd_conf_file
+      @rsyncd_pid_file
+
+      def initialize(sync_name, options)
+        @sync_name = sync_name
+        @options = options
+        # if a custom image is set, apply it
+        if @options.key?('image')
+          @docker_image = @options['image']
+        else
+          @docker_image = 'zluiten/fswatch-rsync'
+        end
+
+        begin
+          Dependencies::Rsync.ensure!
+        rescue StandardError => e
+          say_status 'error', "#{@sync_name} has been configured to reverse sync with rsync, but no rsync binary available", :red
+          say_status 'error', e.message, :red
+          exit 1
+        end
+
+        container_name = get_container_name
+
+        temp = Dir.tmpdir
+        @rsyncd_conf_file = "#{temp}/rsyncd-#{container_name}.conf"
+        @rsyncd_pid_file = "#{temp}/rsyncd-#{container_name}.pid"
+      end
+
+      def start_rsyncd
+        File.open(@rsyncd_conf_file, "w") do |file|
+          file.write(<<NOW
+pid file = #{@rsyncd_pid_file}
+port = #{@options['sync_host_port']}
+reverse lookup = no
+munge symlinks = no
+use chroot = no
+
+[#{@sync_name}]
+hosts deny = *
+hosts allow = 127.0.0.1
+read only = no
+path = #{@options['src']}
+NOW
+          )
+        end
+
+        if File.exist?(@rsyncd_pid_file)
+          stop_rsyncd
+        end
+
+        cmd = "rsync --daemon --config=\"#{@rsyncd_conf_file}\""
+        say_status 'ok', "Starting rsync daemon with #{cmd}", :green
+
+        pid = Process.spawn(cmd)
+        Process.detach(pid)
+      end
+
+      def stop_rsyncd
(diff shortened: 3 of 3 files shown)


netiul:
@EugenMayer What's your fist impression of this feature? I've been using the way it is for quite some time now and although there are a few issues it works quite well.

EugenMayer:
@netiul i am honest, i was not trying it and i am yet not sure how people would need such a sync in a docker-sync development environment. 

For me, docker sync is about development / devops, not some production file transfer - but maybe i have to broaden my view.

could you draw a real-life example on how to use it for what exactly? Maybe pack this as a minimum app into the boilerplate?. Thanks!

michalkleiner:
The first use case that comes to my head for this would node_modules installed inside a node container. Generally you don't need them outside on the host and they are excluded from the sync, but having them on the host 1-way synced from the container would help IDEs to work with them for code hinting etc.

matthew-gill:
Biggest use case for me would be during symfony development. Syncing the cache back from container to host one way only would be really nice!

EugenMayer:
Well thank you guys, interestingly i should have noticed this myself. I am already using exactly that, stacks where i skip docker sync since:
 - i do not sync vendor/node_modules/.m2 - i run `composer install` `npm install` `mvn install` on the container
 - i do not sync the framework files, they are part of the installed symbols in step one
 - i only mount the folder with the 1-5 contrib modules, depending on the framework, i might have to symlink the folder in the container
 - when that startup-script finished, i run something like `copy_symbols.sh` on the host, which then copies vendor/node_modules/.m2 / whatever symbol folder to my local folder so that my IDE has symbol support - which is absolutely critical for me

The latter step is done manually, once running a script. Honestly until now i consider this "ok" due to:
 - one does rarely change dependencies, so rerunning `composer/npm/mvn install` is rare, so resyncing the symbol folders is a rare task
 - so running all this like that

start.sh
```bash
docker-compose up -d
docker exec app composer install
./sync_symbolsh.sh
```

sync_symbolsh.sh
```bash
rm -fr vendor
docker cp app:/var/www/vendor ./
```

----
so that said until know i had the same setup you all mention, just using manual scripts since this is not "code sync", syncing that once a week seems to be already good enough mostly.

Still i understand, that can be automated with docker-sync .. and with little costs:

pros:
 - sync runs entirely on a linux filesystem with linux filewatchers -> no hickups, no pain
 - far less CPU usage then using the inverse setup, less hickups

cons:
 - watching a nodes folder for changes means watchen 30.000 files, that still harms the CPU (for some amount)

---

So my question is, does the pro really overweight the con for a an actual sync which happens so rarely? 

Thank you for sharing your stories!

matthew-gill:
Your setup sounds similar to mine. I sync my src files, but exclude PHP vendor and node modules from syncing. I run a script every now and again to do a docker copy of those  third party dependencies back to host.

My only use case would be for symfony with cache writes only performed on the container. Having the cache up to date on the host is useful for debugging (also the PHPStorm module for symfony uses the cache files).

I'd say I'm still keen for this feature!

michalkleiner:
We actually change npm dependencies quite often when developing on different branches with different dev deps and only merge them once the feature is finished, so part of our workflow, after switching branches, is to run `npm install` to get what the latest lock file prescribes. Potentially same for PHP deps installed via composer.
1-way sync from the container to the host would help us and if it was driven by FS events rather than by watching files, the CPU footprint might be quite small I'd think. Our host env is MacOS.

netiul:
@EugenMayer 
> For me, docker sync is about development / devops, not some production file transfer - but maybe i have to broaden my view.

Don't worry, I used it for development only as well. I wouldn't consider this for production at all.

> could you draw a real-life example on how to use it for what exactly? Maybe pack this as a minimum app into the boilerplate?. Thanks!

I noticed nobody tried the code in this pull request yet, so I guess it's not very clear how it works. I created some real life example using the Symfony Demo app. I pushed it here: https://github.com/netiul/docker-sync-reverse-rsync-symfony-boilerplate

The example mounts the code using the native_osx strategy for syncing, but excludes the `var`, `node_modules` and `vendor` folders. Those folders are separately synced using this `reverse_rsync` strategy. The [README](https://github.com/netiul/docker-sync-reverse-rsync-symfony-boilerplate/blob/master/README.md) contains setup instructions to see it all in action.

A few issues I encouter:

 * One is that with much changes, like a fresh `composer install` or `npm install` the file system event queue will overflow, causing a delay in syncing. How much delay depends, but is max a minute, most of the time less. You can see it happening in this screen recording: https://youtu.be/rgsTsIPaqBM. I find the delay not much of a problem, because these changes are not required to be available with seconds.

 * Another issue is that event queue overflows happen without the user actual performing changes in the folders. I'm not sure why this is happening, but looks like is one reason the strategy is CPU intensive. This needs some more investigation.

> cons:
> 
>  *   watching a nodes folder for changes means watchen 30.000 files, that still harms the CPU (for some amount)

This is true. I've seen in happening with the changes I propose, but this might be less noticeable on smaller projects that I've been using it. That's why I'm curious about other users trying the code.
It also might be less issue if I can fix the occurrence of the sudden event queue overflows. (See previous point.)

netiul:
I'm closing this branch because I'm not happy with the solution as it is and because there's no feedback.

EugenMayer:
@netiul thats a pitty, because i think there has been a really fruitful discussion here about the use cases and that is def. worth it implementing it.

I understand you might be frustrated that there is not a lot of feedback, but i am just too busy with other things and rely on more people jumping it, especially the ones actually going to use it.

Maybe you reconsider your decision. Are there any technical reasons you had to ditch this plan?

netiul:
@EugenMayer I guess I can reopen for the discussion that's already going on here. Altough the issue would be a better place for it.

The reason I'm not happy with the solution is that it's just not performing well enough. When using this `reverse-rsync` strategy in a setup as mentioned in the top post I noticed that it starts to affect the `native-osx` strategy container syncing the actual dev code.

Even when setting the soft and hard open-file limits of on the host machine to 500,000 and 1,000,000 respectively.

EugenMayer:
@netiul thank you for the elaborated answer - i had similar issues / results and that is why i do not use anything with those "watchers".

The issue is not native_osx, but rather OSXFS and the way the FS events stack/propaget. The more events you feed in there the more you get into trouble, be it file-number related or just "more syncs".

Let me know what we should do with this branch - or anybody else. If nobody is really interested in this, i would rather not merge it, since this means there is no maintenance for this

EugenMayer:
understood that this is actually problematic and might just not to be a good, solid way to solve this kind of issue.

Closing the PR now, even though i see so much work already done, as i see also the author went away from it

PR #604, opened 2018-09-06
Na wait for rsync daemon

nalbury-handy:
During start up, our team has been getting:
```
rsync error: error in rsync protocol data stream (code 12) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-52/rsync/io.c(453) [receiver=2.6.9]
```

In our testing we found that it took approx 2-4s for the sync container to not only be running, but for the rsync daemon to be listening on the desired port. The `start_container` method in rsync.rb seemed to handle most of these start delays with a `sleep 3` but if the rsync daemon is still not available after 3s, the initial sync is attempted anyway, resulting in the above error.

The connection to the rsync daemon can be tested with the `--dry-run` flag. So I've attempted a `health_check` method that will use this call to wait until rsync is listening on the desired port and container IP.

This `health_check` method will wait until the` rsync --dry-run...` command returns a non empty string. It will only try 10times to prevent an infinite loop, and any issues syncing after that are handled normally by the `sync` method.

Thanks!

### DIFF
--- a/lib/docker-sync/sync_strategy/rsync.rb
+++ b/lib/docker-sync/sync_strategy/rsync.rb
@@ -41,6 +41,7 @@ module DockerSync
         say_status 'command', cmd, :white if @options['verbose']
 
         out = `#{cmd}`
+        
         if $?.exitstatus > 0
           say_status 'error', "Error starting sync, exit code #{$?.exitstatus}", :red
           say_status 'message', out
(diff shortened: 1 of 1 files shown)


EugenMayer:
that's a pretty good implementation and finding. let me review it roughly and merge it later on. very valuable, thanks!

EugenMayer:
i think one of the issue here might be, that AFAIR `sync_host_ip` is nowdays optional : https://github.com/EugenMayer/docker-sync-boilerplate/blob/master/rsync/docker-sync.yml#L8

i see that we allow `auto` there but i cannot find where we potentially expand it. Did you try it with auto?

nalbury-handy:
Yea in our developer toolkit we don't explicitly set sync_host_ip and it defaulted to 127.0.0.1, which worked ask expected.

EugenMayer:
great thanks! Good to merge?

nalbury-handy:
Yup, thanks!!!

EugenMayer:
I have to thank! Thanks for the contribution.

Issue #621, opened 2018-11-28
`docker-sync` exists with non-zero exit code for properly started project


xulsitatirev:
!!!! Important hint !!!!!

> Please do not create issues for [docker](https://docs.docker.com/) or [docker-compose](https://docs.docker.com/compose/reference) related questions. Ensure you understand the tools properly and consult their documentation first. Please understand what [docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/) / [docker-compose volumes](https://docs.docker.com/compose/compose-file/compose-file-v2/#specifying-byte-values)

> Please be sure to have a look at our [documentation](https://github.com/EugenMayer/docker-sync/wiki) too, especially [configuration](https://github.com/EugenMayer/docker-sync/wiki/2.-Configuration) and our starting point the [boilerplate](https://github.com/EugenMayer/docker-sync-boilerplate)

### Error/Feature Requestion/Docs
* Consitency; `docker-sync` exists with non-zero exit-code for properly started project.
Either something is a warning or an error.

* It is super annoying while scripting.
It is pretty common pattern to type:
```
make start
```
many times, and prepare `docker-sync` for the project.

### Docker Driver
```
Docker Destop
```

### Sync strategy
`unison`

### your docker-sync.yml
```
---
version: '2'
syncs:
  main-app:
    src: ${APP_PATH}
    sync_excludes: [ '.gitignore', '.idea', 'docker-entrypoint.sh']
```

### OS
OSX -- Mojave

PR #622, opened 2018-11-28
Fix exit-code

xulsitatirev:
Closes: #621 -- `docker-sync` exists with non-zero exit code for properly started project


EugenMayer:
I am not sure we won't to treat this this indepotent, could you elaborate your idea/case?

EugenMayer:
sorry just seen your issue. to be honest I think you are right about this, since it is already running. please change the warning to an OK then, color white. thank you for the contribution!

xulsitatirev:
@EugenMayer  Enjoy!

EugenMayer:
Looks good to me, thanks!

PR #630, opened 2019-01-15
WIP: Fix launching of docker-sync via Linux and WSL

### LINKED ISSUE #609: Windows: docker-sync >= 0.5.8 unison-fsmonitor installed not using brew
Error/Docs
during docker-sync start, there are some prompts that doesn't work and it asks for brew installation 
```
docker-sync start
sh: 1: brew: not found
Uninstall legacy unison-fsmonitor (unox)? (y/N) y
     command  sudo rm -f /usr/local/bin/unison-fsmonitor
Install eugenmayer/dockersync/unox for you? [y/N] y
     command  apt-get install -y eugenmayer/dockersync/unox
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
       error  shopsys-framework-sync has been configured to sync with unison, but no unison available
       error  Failed to install required package. Please install it manually and try again.
```
Is there way how to install brew in WSL with functional unison-fsmonitor ?

### Docker Driver
Windows docker 18.06.1-ce

### Sync strategy
unison

### your docker-sync.yml
based on [Wiki](https://github.com/EugenMayer/docker-sync/wiki/docker-sync-on-Windows#fyi---an-example-of-a-docker-syncyml-file)
### OS
Windows 10 Pro 1803

boris-brtan:
fixes #609

### DIFF
--- a/lib/docker-sync/dependencies/unox.rb
+++ b/lib/docker-sync/dependencies/unox.rb
@@ -14,7 +14,7 @@ module DockerSync
       def self.available?
         return @available if defined? @available
         cmd = 'brew list unox 2>&1 > /dev/null'
-        @available = defined?(Bundler) ? Bundler.clean_system(cmd) : system(cmd)
+        @available = Environment.linux? or defined?(Bundler) ? Bundler.clean_system(cmd) : system(cmd)
       end
 
       def self.ensure!


EugenMayer:
Thank you for your work!

Eventhough this fixes the issue, i masks the cause. The point is, it should never get to the point to check for unox under Linux ( und thus WSL ) .. so we have to track back where that cause is, i suppose.

boris-brtan:
yep, that is why i put there WIP afterwards. Is there any debug tool? for instance some plugin for VScode that i could use?

EugenMayer:
i just released a 0.5.10-beta1 - could you verify this fixes your issue? https://rubygems.org/gems/docker-sync/versions/0.5.10.pre.beta1

donnykurnia:
I have made improvement for this hotfix in #636 

I think the reason this changes break on mac, because the usage of `or`. When `defined?(Bundler)` result is `nil` as it should be, the `@available` become false, even when the result of `system(cmd)` is true. At least, this is what I got when tracing the cause of the bugs.

So, in my pull request, brew checking only run on `mac` environment, else just set it to true.

PR #634, opened 2019-01-30
Fix logic for detecting brew unox installation

silverjam:
On my system docker-sync would fail to detect that unox was installed via brew without this fix-up.

Things I can think of that might be related: the system went through a recent Mojave upgrade (2 months) ago, more recently I did a mass "brew upgrade" to resolve an unrelated problem.

However most recently docker-sync prompted me to ugprade from 0.5.9 and 0.5.10-- after the upgrade this problem surfaced.  The problem didn't exist after the Mojave upgrade, but I'm unsure if the problem existed post "brew upgrade".

This PR seems to fix the issue, not sure if the logic is the same since I don't know Ruby.

### DIFF
--- a/lib/docker-sync/dependencies/unox.rb
+++ b/lib/docker-sync/dependencies/unox.rb
@@ -14,9 +14,16 @@ module DockerSync
       def self.available?
         return @available if defined? @available
         cmd = 'brew list unox 2>&1 > /dev/null'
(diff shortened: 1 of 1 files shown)


d4rky-pl:
Please merge this as soon as possible

boris-brtan:
can you try this one ? https://github.com/EugenMayer/docker-sync/pull/636

EugenMayer:
going for  #636 for now

PR #636, opened 2019-01-31
brew is mac specific tools

### LINKED ISSUE #635: unox checking always failed on 0.5.10
!!!! Important hint !!!!!

> Please do not create issues for [docker](https://docs.docker.com/) or [docker-compose](https://docs.docker.com/compose/reference) related questions. Ensure you understand the tools properly and consult their documentation first. Please understand what [docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/) / [docker-compose volumes](https://docs.docker.com/compose/compose-file/compose-file-v2/#specifying-byte-values)

> Please be sure to have a look at our [documentation](https://github.com/EugenMayer/docker-sync/wiki) too, especially [configuration](https://github.com/EugenMayer/docker-sync/wiki/2.-Configuration) and our starting point the [boilerplate](https://github.com/EugenMayer/docker-sync-boilerplate)

Error

### Docker Driver
docker-machine+vbox

### Sync strategy
unison

### OS
OSX

In the version `0.5.10` of docker-sync, there are checking whether unox is installed via brew or not. But whenever I run docker-sync, it will complaint 

```
```

or after I do

```
brew reinstall unox
```

now it complaint

```
```

When I check the history, I saw this commit:
https://github.com/EugenMayer/docker-sync/commit/cbf2758567e834b5e188f0ba5974e0a973ce10ab

What bugging me is that brew checking should only affect mac user, so the checking condition should be

`@available = Environment.mac? && (defined?(Bundler) ? Bundler.clean_system(cmd) : system(cmd))`

I try this in my macbook and now docker-sync can run.

I'll add new pull request for this.

donnykurnia:
I think that the checking should only run on mac. I tried this on my MacBook and now the docker-sync can run.

Fixed #635

### DIFF
--- a/lib/docker-sync/dependencies/unox.rb
+++ b/lib/docker-sync/dependencies/unox.rb
@@ -16,7 +16,7 @@ module DockerSync
         cmd = 'brew list unox 2>&1 > /dev/null'
         # TODO: Environment.linux?  was just a hotfix for something we did not dig deeper into, as we have should
         # see https://github.com/EugenMayer/docker-sync/pull/630
-        @available = Environment.linux? or defined?(Bundler) ? Bundler.clean_system(cmd) : system(cmd)
+        @available = Environment.mac? && (defined?(Bundler) ? Bundler.clean_system(cmd) : system(cmd))
       end
 
       def self.ensure!
--- a/lib/docker-sync/dependencies/unox.rb
+++ b/lib/docker-sync/dependencies/unox.rb
@@ -16,7 +16,12 @@ module DockerSync
         cmd = 'brew list unox 2>&1 > /dev/null'
         # TODO: Environment.linux?  was just a hotfix for something we did not dig deeper into, as we have should
         # see https://github.com/EugenMayer/docker-sync/pull/630
-        @available = Environment.mac? && (defined?(Bundler) ? Bundler.clean_system(cmd) : system(cmd))
+        if Environment.mac?
+          @available = defined?(Bundler) ? Bundler.clean_system(cmd) : system(cmd)
+        else
+          @available = true
+        end
+        @available
       end
 
       def self.ensure!


EugenMayer:
Thank you @donnykurnia .. i guess https://github.com/EugenMayer/docker-sync/pull/634/files is just the same .. a little more elaborate ... would guess this one is a dupe then, am i right?

donnykurnia:
@EugenMayer well, I'm only using docker-sync on my mac. On Linux, I just use normal volume right away. I think if this unox checking also cause a problem in linux, then maybe it's better to do a quick return instead. What do you think @boris-brtan ?

Btw, why the Travis CI build failed?

EugenMayer:
I would consider testing / merging this one and throw out a 0.5.11-beta1 and let the windows/linux and OSX guy test-drive it. From the way i see the logic, what you did makes sense :)

boris-brtan:
yes @donnykurnia, this branch also work with WINDOWS WSL

boris-brtan:
@EugenMayer just two small things, wil new version od docker-sync i was not able to remove old containers via clean so i needed to use `docker rm -f -v ` and maybe [VERSION](https://github.com/EugenMayer/docker-sync/blob/master/VERSION) file should not be 0.5.10 since it is already released, or?

d4rky-pl:
I'm begging you to merge this fix or yank the `0.5.10` version. Explaining daily to people why their docker-sync broke for no apparent reason is really really tiring...

EugenMayer:
Will try to onboard this today, sorry for the delay!

d4rky-pl:
@EugenMayer thank you so much! Please consider [opening a tip jar](https://opencollective.com/) :)

EugenMayer:
Eventhough i merged this one, i decided to jump the gun and try to have a more elaborated, non hotfix way of fixing this

https://github.com/EugenMayer/docker-sync/commit/f07f621fa16fadd6f8a8242e99a724b4e0590237

i double checked were we should check for unox and were we shouldnt, and moved the ckeck all up to the Unison strategy itself - only running it when it is a Mac. 

Tried to cleanup the tests and such.

---

Guys now its time to shine - please thest 0.5.11-beta - all linux and windows and mac users. Report back, lets get this sorted out once and for all! :)

d4rky-pl:
Running it on mac (10.14.2) right now, no issues so far

EugenMayer:
we need at least a windows and/or linux tester

EugenMayer:
We have beta2 on the streets with little adjustments on the actual code but finally fixed tests - thank you to @ignatiusreza !

Could you guys have a run again, we would need a mac and linux tester at least ( surely i tested the former .. but we need non developers to do that .. as always .. it suddenly fails then :) )

d4rky-pl:
For those following just this thread: the update was released ;)

EugenMayer:
Maybe https://twitter.com/DockerSync makes sense for some of you - here i announce betas and releases

Issue #618, opened 2018-11-21
Feature request: specify list of sync endpoints as command-line option


drsharp:
### Error/Feature Request/Docs

I have multiple sync endpoints and sometimes I want to start up some of them, but not all of them. However, it seems that I can only start up all or just one via the `-n <sync-endpoint-name>` flag as per https://github.com/EugenMayer/docker-sync/wiki/2.1-sync-commands

It would be great if I could pass multiple endpoints to that, or call `-n <endpoint1> -n <endpoint2>` or something like that.

This way I can decide which endpoints to start up and can choose more than one but not all of them.

### Docker Driver
d4m

### Sync strategy
default

### your docker-sync.yml
n/a

### OS
OSX 10.14

EugenMayer:
Thanks for contributing!

PR #637, opened 2019-01-31
allow to specific multiple time the sync_name when start

yeuem1vannam:
I'm using the same way as here to detect the sync container so that even the sync daemon is started, it is able to start
https://github.com/EugenMayer/docker-sync/blob/251f39113682cdd74b3bc882a794913ecb666fcb/lib/docker-sync/sync_strategy/native_osx.rb#L83

[asciicast](https://asciinema.org/a/8JvhA89Q75NAS5Z7Q1oWKwOGm)

Two containers are being synced as normal
- app-one-sync
image
- app-two-sync
image

Sorry my local environment cannot run the test so that I made the screencast.

Related to #618

### DIFF
--- a/tasks/sync/sync.thor
+++ b/tasks/sync/sync.thor
@@ -155,8 +155,15 @@ class Sync < Thor
 
       # Check to see if we're already running:
       if daemon_running?
-        say_status 'ok:', 'docker-sync already started for this configuration', :white
-        exit 0
+        should_exit = true
+        unless options[:sync_name].empty?
+          running = `docker ps --filter 'status=running' --filter 'name=#{options[:sync_name]}' --format "{{.Names}}" | grep '^#{options[:sync_name]}$'`
+          should_exit = false if running == ''
+        end
+        if should_exit
+          say_status 'ok:', 'docker-sync already started for this configuration', :white
+          exit 0
+        end
       end
 
       # If we're daemonizing, run a sync first to ensure the containers exist so that a docker-compose up won't fail:


EugenMayer:
A little hesitate to merge this without a test right now ;/

EugenMayer:
Er have the tests up and running again due to the great work of @ignatiusreza in https://github.com/EugenMayer/docker-sync/pull/640 - so i really look forward to merge this in 0.6.0 which should come rather close to the next 0.5.11

I change the scope of 0.6.0 in general, remove quiet some things and move them to 0.7.0 - but your feature should land soon as possible.

We would need some docs too, which are under changes right now https://github.com/EugenMayer/docker-sync/pull/638 so let me merge that one before we deal with that too. 

Thank you very much also for the videos, that helps a lot! 

Question: We should add such a test-scenrio on https://github.com/EugenMayer/docker-sync-boilerplate - so just what you did there? We could use this for feature testing, referencing it in the docs and just showing off what is possible. What are your thoughts?

yeuem1vannam:
@EugenMayer That's great to see the test green again, I will update the PR with tests as soon as possible ( it should be within this weekend )

About the test-scenario, I didn't look into it before, but I think it should be good to have such a scenario there. In my screencast, there are 2 simple ruby apps, so if you think it is good to be one scenarios, I will make a PR to [EugenMayer/docker-sync-boilerplate](https://github.com/EugenMayer/docker-sync-boilerplate)

EugenMayer:
@yeuem1vannam use the concept of tests we already have there, so no real apps, just "`pseudo cat`" which also already shows if sync is working. 

Looking forward to your PR and thank you for the effort!

EugenMayer:
@yeuem1vannam you should merge this branch upstream so tests are already green for you

yeuem1vannam:
@EugenMayer I've merged to upstream's master.

EugenMayer:
Thank you @yeuem1vannam - that looks good already! I will have a final look on that and the we wrap this up

EugenMayer:
thank you a lot!

Issue #695, opened 2019-11-07
docker-sync start when already started raises an exception


anlek:
### Error
When you run `docker-sync start` and the sync is already running, you get an error:
```
tasks/sync/sync.thor:159:in `daemonize': undefined method `empty?' for nil:NilClass (NoMethodError)
```

![image](https://user-images.githubusercontent.com/11052/68409526-34ecbf00-0155-11ea-88d0-cc7473350077.png)

In the earlier version, if you called `docker-sync start` when it was already running, it would respond with **"Sync already running"** (or something like that).

### docker-sync version
0.5.13

### Docker Driver
N/A

### Sync strategy
N/A

### your docker-sync.yml
N/A

### OS
MacOS 10.14.6

martmull:
I have got the same error since this morning

dimonixx:
Me too

Maxooo:
Same behavior here after the Mac OSX Catalina (`10.15.1` - build `19B88`) update.

Deconstrained:
I am convinced I have found the cause of the issue,

The value is expected to default to an empty string (ostensibly) but it's coming through as nil for whatever reason.

PR #698 addresses this by explicitly setting a default.

EugenMayer:
This makes pretty much sense -this was introduced in the last release and we forgot to care about the proper defaults. Thanks for your effortT

Maxooo:
Thank you @Deconstrained for your work and @EugenMayer for the new `v0.5.14` version which works perfectly now. Your investment in this project is well appreciated and makes us win so much time with Docker on Mac every-day. 🙏

EugenMayer:
The main driving force is the community right now, I just take time to coordinate / stuff together. So I forward all the credits to you gents and lads :)

PR #698, opened 2019-11-19
Address issue #695

Deconstrained:
Despite the `:sync_name` argument being configured as a string type, in Mac OS X / a few versions of Ruby reported, the argument's value gets interpreted as nil when not given on the command line.

~This sets the default for the argument to an empty string so that when `empty?` is passed to it, it won't produce `NoMethodError` the way it would for `nil`.~ Setting the default to an empty string causes other issues.

This modifies the condition statement where the exception was raised in issue #695  so that it short-circuits when the value is `nil`.


EugenMayer:
Thank you for your contribution @Deconstrained !

PR #700, opened 2019-12-03
Unison needs to compile with ocaml 4.08.1

rulin132:
Updated Windows installation section to document Unison needing to compile with Ocaml 4.08.1 as well as made it more clear how to determine what version of Ocaml is needed to be install.


EugenMayer:
Thank you a lot, we forgot about this during the update!

Issue #713, opened 2020-02-17
docker-sync restart subcommand


kurarrr:
### Feature Requestion

Hi there.
I often `docker-sync stop && docker-sync start` when some sync issues occur.
Can I add `docker-sync restart` subcommand (or does it exist)?

EugenMayer:
Thank you for working on that one and contributing, merged and ready to be released

PR #714, opened 2020-02-21
add restart command

kurarrr:
https://github.com/EugenMayer/docker-sync/issues/713
I added `docker-sync restart` command.


EugenMayer:
That looks really good already, thank you for the conribution. Could you add some notes to the documentation about this command. I would assume it in 

https://github.com/EugenMayer/docker-sync/blob/master/docs/getting-started/commands.rst

so 

https://docker-sync.readthedocs.io/en/latest/getting-started/commands.html

kurarrr:
Thank you for the check, I forgotten to do that.
I added a short explanation of the command. Please look through that.

PR #715, opened 2020-02-28
Add required version number to example

cabello:
Hi 👋 

Thanks for making and maintaining this project. While trying it for the first time I noticed that the documentation is not using the required version entry in the configuration file.

Hope this is useful for other people.


EugenMayer:
Thanks, good find

PR #716, opened 2020-02-28
Rename default-sync to shortexample-sync

cabello:
Hi 👋 

While reading the documentation I was under the impression that `default-sync` followed some type of hierarchy, for example if I defined `foo-sync` I would expect it to use all settings defined in `foo-sync` plus `default-sync`, like a object merge between the two, that's not the case, it's just the name of the volume. To avoid this potential confusion I am proposing renaming it to `shortexample-sync`.

### DIFF
--- a/docs/getting-started/configuration.rst
+++ b/docs/getting-started/configuration.rst
@@ -59,12 +59,12 @@ References
       project_root: 'pwd'
 
     syncs:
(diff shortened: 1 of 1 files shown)


EugenMayer:
Again, makes a lot of sense, i can def. understand the POV. Thank you!

PR #730, opened 2020-04-23
Fix wrong unison source path in unison strategy. Fixes #728

### LINKED ISSUE #728: Unison strategy does not start
Looks like this sync strategy isn't maintained at all. First, there are issues with installation script - it goes into infinite loop.
```
Uninstall legacy unison-fsmonitor (unox)? (y/N) y
     command  sudo rm -f /usr/local/bin/unison-fsmonitor
Password:
Install eugenmayer/dockersync/unox for you? [y/N] y
     command  brew install eugenmayer/dockersync/unox
Warning: eugenmayer/dockersync/unox 0.2.0_1 is already installed and up-to-date
To reinstall 0.2.0_1, run `brew reinstall unox`
Install eugenmayer/dockersync/unox for you? [y/N] y
     command  brew install eugenmayer/dockersync/unox
Warning: eugenmayer/dockersync/unox 0.2.0_1 is already installed and up-to-date
To reinstall 0.2.0_1, run `brew reinstall unox`
Install eugenmayer/dockersync/unox for you? [y/N]
       error  shop-sync has been configured to sync with unison, but no unison available
       error  Did not install required package. Please install it manually and try again.
```

Following that, there are some issues with ports
```
          ok  Starting unison for sync community-sync
          ok  community-sync container still running, restarting unison in container
     command  docker inspect --format='{{(index (index .NetworkSettings.Ports "5000/tcp") 0).HostPort}}' community-sync
       error  Error getting mapped port, exit code 0
     message  Template parsing error: template: :1:3: executing "" at <index (index .NetworkSettings.Ports "5000/tcp") 0>: error calling index: index of untyped nil
```
Screenshot 2020-04-14 at 23 30 54

### Docker Driver
d4m

### Sync strategy
unison

### your docker-sync.yml
```yml
version: "2"

syncs:
  community-sync: &community
    sync_strategy: unison
    src: ./community
  shop-sync:
    <<: *community
    src: ./shop
```
### OS
MacOS Catalina

ostrolucky:
Unison currently crashes with error "File monitoring helper program not running"
That happens because unox crashes with
```
[unox][DEBUG]: sendCmd: ERROR %5BErrno%202%5D%20No%20such%20file%20or%20directory%3A%20%27/app_sync/%27
```

Indeed, /app_sync is destination path in docker container, not source path on host

It should be like here https://github.com/EugenMayer/docker-sync/blob/2f3b65eaf236955b9a14f363b817f85b392422be/lib/docker-sync/sync_strategy/unison.rb#L120

### DIFF
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -204,7 +204,7 @@ module DockerSync
         say_status 'ok', "starting initial sync of #{container_name}", :white if @options['verbose']
         # wait until container is started, then sync:
         sync_host_port = get_host_port(get_container_name, UNISON_CONTAINER_PORT)
-        cmd = "unison -testserver #{@options['dest']} \"socket://#{@options['sync_host_ip']}:#{sync_host_port}\""
+        cmd = "unison -testserver #{@options['src']} \"socket://#{@options['sync_host_ip']}:#{sync_host_port}\""
         say_status 'command', cmd, :white if @options['verbose']
         attempt = 0
         max_attempt = @options['max_attempt'] || 5


EugenMayer:
I did not look into the patch deeply, but this patch means, that nobody yet can use unison, which is not true. 

Applying this patch though would lead to exactly that, if it is currently working.

I will have to look at the patch more carefully but i do not expect it to be mergeable by the above - you might have a configuration issue on your side rather then we flipped src/dest after the strategy is arround for 3 years already - it's just very unlikely

EugenMayer:
double checked, this line is there since 3 years with the exact same options since the entire strategy was implemented. It's fairly save to say out of this reasons, it cannot be wrong, a lot of people have used unison in the meantime.

This is not the cause of the issue, maybe a symptom

ostrolucky:
It works normally because with CLI version of unison (the one used when doing `brew install unison`), this path doesn't matter. Unison just attempts to connect and immediately disconnects. Docker-sync uses it just to check if server is available.

However when having unison with GUI as was my case, this path becomes important and user gets error messages because of it. So it's wrong path nevertheless. **This patch is important for users in same situation like me, because it allows docker-sync progress further in process, where user gets better error messages (client/server versions mismatch)**

Please don't be so quick to close pull requests, I'm trying to help your project here. Yes, this line worked for majority users. But it worked by accident, path was always wrong.

EugenMayer:
Well your entire comment should have been in here in the first place - you did neither explain why it fixes it, what it fixed, why it worked back in time and what you are doing differently in the first place.

You should have opened an issue for this, explaining what is what, basically https://github.com/EugenMayer/docker-sync/pull/730#issuecomment-618850105 and then open this PR.

Don't through in PRs just like that, so i wont't close them just like that - deal? :)

You just need to be aware how many PR's we had for people fixing _their_ setup, not the general setup, braking the general setup - so i'am always sensible to those kind of commits were potentially working code for **_a lot of people_** is changed, since **_one_** person has issues - i hope you can understand what i mean here.

Thank you for elaborating on this, still need to understand what this parameter means in this regard and why "this path becomes important out of a sudden" - seems like this path should be critical in all cases

ostrolucky:
There is an issue for this - #728. It's linked to in this pull request title. In last comment of that issue I think I might have explained it sufficiently, but indeed at that point and at point of creating this PR I didn't yet know this behaviour is different for GUI unison and CLI unison. Path used there was wrong whole time though nevertheless.

As for why this is a difference between GUI and CLI versions I cannot explain, that would require reading Unison source code which would be quite a feat. But it shouldn't be important either, you just need to realize this behaviour is different between these and can be easily tested with `brew uninstall unison && brew cask install unison && docker-sync-stack start`

ostrolucky:
Hey, anything else in need of clarifying here? Merging this will solve issue #728

EugenMayer:
> As I explained earlier, my assumption is this part of docker-sync code is used only to check if unison server is running, so path for unison shouldn't be critical. It's specified only because otherwise Unison CLI argument validation fails. Docker-sync doesn't expect any synchronization starting at this point. But Unison GUI does some additional things I cannot explain.

I think that one was the missing link i was looking for - why did we not expect the path to be valid until know - but this is just a smoke test and the param is mandatory, so nobody ran into it being the wrong path.

Thank you once again for clarifying. Going to merge this one now

PR #739, opened 2020-05-08
Use COMPOSE_FILE to set default compose file

patsplat:
docker-compose allows the `COMPOSE_FILE` environment variable to control the docker-compose file:

https://docs.docker.com/compose/reference/envvars/#compose_file

This PR updates docker-sync to also use `COMPOSE_FILE`

### DIFF
--- a/lib/docker-sync/compose.rb
+++ b/lib/docker-sync/compose.rb
@@ -8,7 +8,10 @@ class ComposeManager
     @global_options = global_options
 
     ### production docker-compose.yml
-    compose_files = [File.expand_path('docker-compose.yml')]
+    compose_files = DockerSync::Environment.compose_file.map do |f|
+      File.expand_path(f)
+    end
+    
     if @global_options.key?('compose-file-path')
       compose_files = [] # replace
       apply_path_settings(compose_files, @global_options['compose-file-path'], 'compose-file-path')
--- a/lib/docker-sync/environment.rb
+++ b/lib/docker-sync/environment.rb
@@ -19,5 +19,13 @@ module DockerSync
     def self.system(cmd)
       defined?(Bundler) ? Bundler.clean_system(cmd) : Kernel.system(cmd)
     end
+
+    def self.compose_file
+      if ENV['COMPOSE_FILE']
+        ENV['COMPOSE_FILE'].split(/[:;]/)
+      else
+        ["docker-compose.yml"]
+      end
+    end
   end
 end
--- a/spec/lib/docker-sync/environment_spec.rb
+++ b/spec/lib/docker-sync/environment_spec.rb
@@ -1,0 +1,46 @@
+require_relative "../../../lib/docker-sync/environment"
+
+RSpec.describe DockerSync::Environment do
+  describe :compose_file do
+    def mock_compose_env(value)
+      allow(ENV).to receive(:[]).with("COMPOSE_FILE").and_return(value)
+    end
+
+    it "should have the correct default" do
+      mock_compose_env(nil)
+
+      result = DockerSync::Environment.compose_file
+
+      expect(result).to eq(['docker-compose.yml'])
+    end
+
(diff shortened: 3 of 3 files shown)


EugenMayer:

Since we have specific settings in the docker-sync.yaml for that particular purpose, this collides with the current implementation and also is not clear what superseeds the other. This is just a feature on top of the same feature and we have no concept of making them work together.

I suppose this is not a clear go at all

patsplat:
@EugenMayer thanks for the response!

I created this PR because the settings in `docker-sync.yaml` didn't support my purpose.

- Have a large team using `docker-sync`
- Have a new feature which requires new services
- Want to turn that feature on in local dev on an opt-in basis with environment variables

The way the `docker-sync.yaml` config variables work, there didn't seem to be an option for a default of `["docker-compose.yml"]` to become `["docker-compose.yml", "new-feature.yml"]` solely based on ENV config.
- Got errors that the environment variable was missing
- Didn't seem to parse the multiple path in a single line syntax

I considered suggesting making changes to the settings in docker-sync.yaml, but felt aligning with `docker-compose`'s `COMPOSE_FILE` was easier to explain and would be easier to maintain.

patsplat:
This PR updates the default behavior of `docker-sync-stack` to align with the default behavior of `docker-compose`.

The semantics of the explicit `docker-sync` config options remain unchanged.

patsplat:
Note that without this PR, one can still:

```
export COMPOSE_FILE=docker-compose.yml:another-compose.yml
docker-sync start
docker-compose up
```

This PR enables:
```
export COMPOSE_FILE=docker-compose.yml:another-compose.yml
docker-sync-stack start
```

EugenMayer:
Did you consider https://docker-sync.readthedocs.io/en/latest/getting-started/configuration.html 

```yaml
options:
  compose-dev-file-path: 'docker-compose-dev.yml'
```

i know the ENV vars you are talking about and we are using them ourselfs a lot without docker-sync - but docker-sync has no support for them yet. If you implement what you have, people using the above might have a broken configuration since it is overriden.

Please do extensive tests while doing both, using the above and COMPOSE_FILE at the same time, why you need to get the yaml file as "overriding the ENV var", so our old default works.

Thank you for the effort

patsplat:
Yes I had initially tried using those variables.

- If these variables are set, then the files must exist
- If an environment variable is used to set these variables, then the env variable must exist

I was unable to figure how to load a different config solely via an optional env variable.

I found myself with two options:
- add support for default values in the env substitution logic
- add support for using COMPOSE_FILE as an alternative config method

patsplat:
By extensive tests are you suggesting trying the config variables?

Or are there more edge cases to consider prior to accepting this PR?

EugenMayer:
Hello @patsplat, sorry for me not answering for so long. I would like to take on this one to if possible, incorporate your already valueable work.

Do you use this variant in your office right now ( daily ) and if yes, what kind of setup are you using. Could make a little boilerplate as an example?

patsplat:
@eugenmayer thank you for the response.

The use case was using COMPOSE_FILE to control feature flags.

For example `COMPOSE_FILE=...:docker-compose-ssr.yaml` where docker-compose-ssr.yaml contains services and settings to enable server side rendering.

In this way, most developers would see no disruption while the new batch of services is developed.

patsplat:
Our application has already evolved uses for `docker-compose-dev.yml`

I was trying to avoid the noise of "paste this large batch of new services into `docker-compose-dev.yml`", at least until the feature was more stable.

EugenMayer:
hello @patsplat - are you still using this feature?

patsplat:
We'd actually migrated away from docker-sync.  I don't mind shaping up this PR, but no this branch is not being actively used.

EugenMayer:
nevermind, probably not worth the effort then. Thank you for the feedback!

PR #740, opened 2020-05-09
Detect docker host OS

dukaev:
👋 @EugenMayer. Thank you for this app

I have an external machine in the docker-machine. docker-sync works with it only if specify the address:
```
sync_host_ip = 91.117.201.122
```
`sync_host_ip = 'auto'` doesn't work because of my docker host OS is `Ubuntu 18.04.4 LTS`
I have a `macOS` and `docker toolbox` installed there.

My PR checks that the OS docker exists. Not only `Boot2Docker`

I will be glad to answer your questions

### DIFF
--- a/spec/lib/docker-sync/dependencies/docker_driver_spec.rb
+++ b/spec/lib/docker-sync/dependencies/docker_driver_spec.rb
@@ -60,7 +60,7 @@ RSpec.describe DockerSync::Dependencies::Docker::Driver do
 
     it 'checks if Docker is running in Boot2Docker' do
       subject
-      expect(DockerSync::Environment).to have_received(:system).with('docker info | grep -q "Operating System: Boot2Docker"')
+      expect(DockerSync::Environment).to have_received(:system).with('docker info | grep -q -E "Operating System: {1,}"')
     end
 
     it 'is memoized' do
--- a/lib/docker-sync/dependencies/docker_driver.rb
+++ b/lib/docker-sync/dependencies/docker_driver.rb
@@ -12,7 +12,7 @@ module DockerSync
           return false unless Environment.mac? || Environment.freebsd?
           return false unless find_executable0('docker-machine')
           return @docker_toolbox if defined? @docker_toolbox
-          @docker_toolbox = Environment.system('docker info | grep -q "Operating System: Boot2Docker"')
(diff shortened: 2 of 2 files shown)


EugenMayer:
Did you test this under mac?  This detection was pretty hard to nail down in the past

I cannot see how this would not detect a docker-for mac as "system" too

dukaev:
@EugenMayer yes, I did. Do you have a mac? Check these steps:
1. Add docker external docker (from http://hetzner.com/ for example) to your docker-machine
```
docker-machine create \
    --driver generic \
    --generic-ip-address=$IP_ADDRESS \
    --generic-ssh-user $USERNAME \
    --generic-ssh-key ~/.ssh/$PRIVATE_KEY \
    default

```

2. Set it by default
```
eval $(docker-machine env default)
```

3. Use `sync_host_ip` by default value.

dukaev:
image

EugenMayer:
Apprently currently i do no longer run macOS for development, using appveyor for macOS based tests will be one of my goals for docker-sync.

I understood your point, the question is, what does a d4m installation return in `Operating System` - i suspect it also would return somewhat an alpine linux flavor.

Why this is important is
 - with your patch, `.docker_toolbox?` would return true inside a docker-4-mac installation
 - without your patch i wont

This is a IMHO - it is what i remember, i cannot test this right now

dukaev:
@EugenMayer how can I help fix this?

EugenMayer:
Hello @dukaev any way we can move this forward?

dukaev:
Hello @EugenMayer. Do you have any ideas?

spec/lib/docker-sync/dependencies/docker_driver_spec.rb
```diff
@@ -60,7 +60,7 @@
 
     it 'checks if Docker is running in Boot2Docker' do
       subject
-      expect(DockerSync::Environment).to have_received(:system).with('docker info | grep -q "Operating System: Boot2Docker"')
+      expect(DockerSync::Environment).to have_received(:system).with('docker info | grep -q -E "Operating System: {1,}"')
```
grepsedawk, line 63: The expectation doesn't seem to match the test description after this change

Issue #39, opened 2016-07-12
[idea] Rework configuration reference to explain each option


EugenMayer:
Instead of showing a annotated configuration, rather split into global options and syncs option and explain each value as:
- meaning
- default
- possible values / allowed values
- optional / mandatory

anybody willing to help here? Be bold https://github.com/EugenMayer/docker-sync/wiki/2.-Configuration

everybody has write access

thizzle:
Would a `docker-sync init` task that auto-generates a fully-documented sample `docker-sync.yml` file be useful? Similar to how `vagrant init` works.

EugenMayer:
That Sounds like a realy good idea. Go forward doing that happy to merge such in

henscu:
Thanks to docker-sync I've just got my Bedrock WordPress Docker Compose (V3) setup working and would be glad to add a section to the wiki for this it would help many WordPress developers.

How does updating the wiki work? I just edit it and commit?

EugenMayer:
You Fan freely create neu Wiki Pages, just be bold!

codycraven:
I'd really like to see a good description of each of the sync options, then breaking each configuration explanation into its own page.

This would allow the user to determine which option will be best for their use, then go access example and details strictly addresses their sync option. For example, I don't know right now if I should define _sync_host_port_ when using native_osx because the wiki mentions not using it for unison, but also that it is for rsync.

EugenMayer:
@unnawut any chance you would pick this up. This would be so great :)

unnawut:
Note that I haven't forgotten this issue but havn't got time to get to it yet. Will be back!

EugenMayer:
@unnawut you have done so much already, so i am already more then happy! If you can take this on anytime, this is just over the top :)

Thanks!

cweiske:
This can be closed.

EugenMayer:
Right, thanks for the follow up and thank you for all the work!

PR #744, opened 2020-07-09
Property documentation: Rework configuration reference

cweiske:
Resolves: https://github.com/EugenMayer/docker-sync/issues/39

### DIFF
--- a/docs/getting-started/configuration.rst
+++ b/docs/getting-started/configuration.rst
@@ -40,6 +40,12 @@ General options
 ---------------
 Configuration options below the top-level ``options`` key.
 
+.. code-block:: yaml
+
+    version: "2"
+    options:
+      verbose: true
+
 
 cli_mode
 ^^^^^^^^
@@ -153,6 +159,14 @@ possible values:
 Synchronization options
 -----------------------
 Configuration options for a synchronzation definition below the top-level ``syncs``.
+
+.. code-block:: yaml
+
+    version: "2"
+    syncs:
+      appcode-native-osx-sync:
+        src: './app'
+        sync_excludes: ['ignored_folder', '.ignored_dot_folder']
 
 
 host_disk_mount_mode
--- a/docs/getting-started/configuration.rst
+++ b/docs/getting-started/configuration.rst
@@ -21,124 +21,415 @@ Below are all the available options, simple examples can be found in the docker-
 
 .. _docker-sync-boilerplate: https://github.com/EugenMayer/docker-sync-boilerplate
 
-References
-----------
 
-.. caution::
-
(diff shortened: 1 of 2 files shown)


EugenMayer:
wow that is quiet some chunk of work, thank you so much. right now trying to understand how to see this pr on readthedocs to see how  this would view there..

cweiske:
I used a local sphinx installation to render it

EugenMayer:
could you just do an additional commit onto this pr, i just activated pr builds with readthedocs but it needs to be retriggered

docs/getting-started/configuration.rst
```diff
@@ -21,124 +21,415 @@ Below are all the available options, simple examples can be found in the docker-
...
+        sync_excludes: ['ignored_folder', '.ignored_dot_folder']
 
-      # optional, default: docker-compose-dev.yml if you like, you can set a custom location (path) of your compose file. Do not set it, if you do not want to use it at all
 
-      # if its there, it gets used, if you name it explicitly, it HAS to exist
-      # HINT: you can also use this as an array to define several compose files to include. Order is important!
-      compose-dev-file-path: 'docker-compose-dev.yml'
+General options
+---------------
+Configuration options below the top-level ``options`` key.
```
EugenMayer, line 41: we should add a visual example here, what do you think?

docs/getting-started/configuration.rst
```diff
@@ -21,124 +21,415 @@ Below are all the available options, simple examples can be found in the docker-
...
 
-      # optional, default: pwd, root directory to be used when transforming sync src into absolute path, accepted values: pwd (current working directory), config_path (the directory where docker-sync.yml is found)
-      project_root: 'pwd'
+default:
+  ``auto``
+
+possible values:
+  - ``docker-sync``
+  - ``thor``
+  - ``auto``
```
EugenMayer, line 58: maybe we should use a different format here, e.g. a table with 2 columns,first col for the possible values, the second is the default? what is your opinion?

cweiske: I use the "possible values" to explain the option sometimes, so having them separate is helpful IMO

docs/getting-started/configuration.rst
cweiske: Done.

docs/getting-started/configuration.rst
EugenMayer: sure, its more about table vs list

EugenMayer:
Seems like the readthedocs PR build works either for only local branches or for PRs newly created. If you like, just recreate your PR - mine seems to work https://github.com/EugenMayer/docker-sync/pull/745

I added a docker based variant to work on the docs in `docs/README.md`

cweiske:
Continuing in #746.

PR #752, opened 2020-09-11
Update configuration.rst

flora8984461:
Fix missing a "t" in the config of max_attempt, I tried many times when doing my config then suddenly noticed it.


EugenMayer:
Nice catch, thank you!

PR #753, opened 2020-09-20
Update link in message regarding `dest` removal

lidimayra:
When a value for `dest` is set, a message is displayed informing that it has been removed. This message contains a link pointing to wiki as a reference. This link is outdated though.

This commit replaces the outdated link with a link to the same section in the docs that are currently being used (readthedocs).


EugenMayer:
Thank you for catching that one

PR #782, opened 2021-09-30
Fix grammar in dependencies/unox.rb then -> than

grepsedawk:
In unox.rb, the error raised if the device is mac states:
'Unox cannot be installed on platforms other than MacOS'
but I think the then should be than

Source: https://www.merriam-webster.com/words-at-play/when-to-use-then-and-than


EugenMayer (approved):
Thanks!

Issue #788, opened 2021-11-10
Replace docker-compose gem with direct shell invocations


EugenMayer:
https://github.com/EugenMayer/docker-sync/blob/master/docker-sync.gemspec#L16 is outdated and is no  longer developed.

Using the wrapper introduces issues and limits the flexibility and features we can use, e.g. like #787 but also others

Replacing the gem with direct shell invocations would mean

- replace all usages with a direct cli invocation
- ensure we stay cross-platform (i guess as long as our invoke is x-platform, the docker-compose binary is pretty much the same)

I think it is not a small effort but also not a huge one. On the other hand, it could lead to quiet some handy features in the future.

Anyone opting in to go for that one?

EugenMayer:
will be closed with #795

Issue #787, opened 2021-11-09
Passing Through Options to `docker-compose`


KidA78:
### Error/Feature Requestion/Docs
I've created a profile in `docker-compose` that I'd like to be able to turn on when using `docker-sync-stack start` 
How might I be able to do this?

### OS
Mac OS Catalina

EugenMayer:
I would say, this is not possible. we are using an abstraction to trigger docker-compose and that abstraction does not support that. created #788 but be sure to notice, i will not work on that myself, someone else will need to step in. Maybe you?

EugenMayer:
will be closed with #795

PR #795, opened 2022-02-20
Remove dependency on docker-compose gem; support Ruby 3.x

### LINKED ISSUE #777: Ruby 3: `gem install docker-sync` installs version 0.0.11
Attempting to install the `docker-sync` gem in a Ruby 3.x environment results in the v0.0.11 version of docker-sync being installed.

Attempting to install higher versions of the `docker-sync` gem results in the following error:
```
❯ gem install docker-sync -v 0.6.0
ERROR:  While executing gem ... (NoMethodError)
    undefined method `request' for nil:NilClass
```

To investigate, I cloned the `docker-sync` git repository and attempted to run `bundle update` within:
```
❯ bundle update
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    Ruby

    docker-sync was resolved to 0.6.0, which depends on
      docker-compose (~> 1.1, >= 1.1.7) was resolved to 1.1.12, which depends on
        Ruby (~> 2.0)

    docker-sync was resolved to 0.6.0, which depends on
      Ruby (>= 2.4)

    rspec was resolved to 3.10.0, which depends on
      rspec-core (~> 3.10.0) was resolved to 3.10.1, which depends on
        Ruby (>= 1.8.7)

    docker-sync was resolved to 0.6.0, which depends on
      thor (~> 1.0, >= 1.0.0) was resolved to 1.1.0, which depends on
        Ruby (>= 2.0.0)

Bundler could not find compatible versions for gem "docker-sync":
  In Gemfile:
    docker-sync
```
which reveals the underlying issue lies with the `docker-compose` gem, which depends specifically on Ruby 2.x.

Going further down the rabbit hole, `docker-compose` depends on `backtick`, which again depends on Ruby 2.x.

takahashim:
Hi,
I'm new to this gem, but my team is using docker-sync for daily development. Thank you very much for this useful gem!

To fix #777, I tried to do #788, took only the minimal functionality from docker-compose and backticks gems and combined it into the `DockerSync::DockerComposeSession` and `DockerSync::Command` classes.

These classes are limited in functionality and less flexible (ex. arguments of `Command.run` is just Array of String, not Hash). However, it is rather desirable in terms of stability, and we can add more functions if needed (to solve #787 or other issues).


EugenMayer:
This sounds great already! I guess we have to be somewhat careful to understand how the current options could parametrise docker-compose so we can keep the same level at least

The reason to get rid of the gem was

A) more flexibility and native/direct options support
B) docker-compose configuration file overloading
C) prepare ruby 3 support

I think the most sensible/ wanted part was b)

Did you already scan the configuration for options we need to pipe?

Thank you so much for working on this!

takahashim:
Thank you for your reply!  I'm glad you like it.

The docker-compose CLI options that this PR supports may still be insufficient, but we currently support the following options:

### docker-compose up

* `docker-compose up` (uses default `docker-compose.yml`)
* `docker-compose --file=foo.yml up`
* `docker-compose --file=foo.yml --file=bar.yml (... or more --file options) up`
* `docker-compose --file=foo.yml --file=bar.yml up --build`

### docker-compose stop

* `docker-compose stop`
* `docker-compose --file=foo.yml --file=bar.yml (...) stop`

### docker-compose down

* `docker-compose down`
* `docker-compose --file=foo.yml --file=bar.yml (...) down`

I've modified the spec and added a few more.
I think this makes it easier to understand what the `DockerComposeSession` class currently supports.

https://github.com/EugenMayer/docker-sync/blob/2b19fe696afb7a7aa311c8d412f3b24d04e3e547/spec/lib/docker-sync/docker_compose_session_spec.rb#L22-L93

> A) more flexibility and native/direct options support

For this point, what `DockerComposeSession` actually doing is just pass command arguments as an Array of String to the `Kernel#spawn` method; so it's direct enough, I guess.

> B) docker-compose configuration file overloading

As I wrote above, the (multiple) `--file` options of docker-compose CLI is supported.

takahashim:
By the way, I'm using `--file=foo.yml` style for file options to match the current behavior (I think), but should I use `-f foo.yml` style with short options?
(Because the manual of docker-compose was the latter.)

EugenMayer:
Really nice.

A really nice take-away from this PR would be the support of .env and COMPOSE_FILE in there, so one can have customer overrides per .env for docker-compose. This was relevant for macos and linux teams, where the latter wanted to use docker-sync, but basically as a dummy - and using an .env file for that makes a lot of sense.

I will review the PR ASAP so we can move on with the issue. I think you are basically tackling several issues and features with this single PR at the same time.

Thank you for your work!

EugenMayer:
@takahashim sorry for the huge delay on this one. I would like to pick this one up since i think your work is really valueable and a lot of effort has been put into that already.

I'am sure you are running your variant for a while now, anything suprising we need to consider? Are you still using the spin off in production?

EugenMayer:
@takahashim is there any chance to get you motivated or any type of help i could provide so you finish this really valuable effort? Let me know!

EugenMayer:
@takahashim i re-rolled your branch and merged downstream in #803 since - let me know if you want to continue here, happy to close the other branch in favour of your work here. This or that way, all the credits for the work are yours.

PR #811, opened 2022-08-25
Fix unison image architecture

yoseph1998:
The current docker-sync will always pull the AMD unison image, even when running on ARM systems.

To fix this we can set the default unison image tag based on the architecture of the platform with the `RUBY_PLATFORM` constant.

### DIFF
--- a/lib/docker-sync/sync_strategy/native_osx.rb
+++ b/lib/docker-sync/sync_strategy/native_osx.rb
@@ -23,7 +23,11 @@ module DockerSync
         if @options.key?('image')
           @docker_image = @options['image']
         else
-          @docker_image = 'eugenmayer/unison:2.51.3-4.12.0-AMD64'
+          if RUBY_PLATFORM.include? "arm"
+            @docker_image = 'eugenmayer/unison:2.51.3-4.12.0-ARM64'
+          else
+            @docker_image = 'eugenmayer/unison:2.51.3-4.12.0-AMD64'
+          end
         end
 
         begin
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -22,7 +22,11 @@ module DockerSync
         @docker_image = if @options.key?('image')
                           @options['image']
                         else
-                          'eugenmayer/unison:2.51.3-4.12.0-AMD64'
+                          if RUBY_PLATFORM.include? "arm"
+                            'eugenmayer/unison:2.51.3-4.12.0-ARM64'
+                          else
+                            'eugenmayer/unison:2.51.3-4.12.0-AMD64'
+                          end
                         end
         begin
           Dependencies::Unison.ensure!


lib/docker-sync/sync_strategy/native_osx.rb
```diff
@@ -23,7 +23,11 @@ def initialize(sync_name, options)
         if @options.key?('image')
           @docker_image = @options['image']
         else
-          @docker_image = 'eugenmayer/unison:2.51.3-4.12.0-AMD64'
+          if RUBY_PLATFORM.include? "arm"
```
EugenMayer, line 26: we should put this detection into a tool method that is more semantic and can be reused on both places.

i assume https://github.com/EugenMayer/docker-sync/blob/master/lib/docker-sync/environment.rb makes most  sense?

EugenMayer (approved):
thank you for the contribution. just minors to adjust.

Did you test it in both strategies e2e?

blackjid:
Why the issue is solved in docker-sync? I think it would be better to let docker to pick the correct image from the registry based on the registry and using manifests. For that, docker-sync should default to an image that doesn't specify the architecture. like `eugenmayer/unison:2.51.3-4.12.0`

I was playing a bit with the travis configuration when building the image, it builds both amd64 and arm64 and bundle them with a manifest on the same image.

https://github.com/EugenMayer/docker-image-unison/compare/master...blackjid:docker-image-unison:test_amd
That is branch I was working on, it works. The only issue is that it uses the same job to build both, so it need an arm64 emulator (qemu) to compile the arm64 version, and it's painfully slow..

Maybe it doesn't matter it is slow because you just compile the image once in a while.

The way to fix it is to run travis on different jobs each on the correct arch (like the current travis config is doing) and after that create and push the manifest to the registry to bundle both images in one

yoseph1998:
> Why the issue is solved in docker-sync? I think it would be better to let docker to pick the correct image from the registry based on the registry and using manifests. For that, docker-sync should default to an image that doesn't specify the architecture. like `eugenmayer/unison:2.51.3-4.12.0`
> 
> I was playing a bit with the travis configuration when building the image, it builds both amd64 and arm64 and bundle them with a manifest on the same image.
> 
> [EugenMayer/docker-image-unison@master...blackjid:docker-image-unison:test_amd](https://github.com/EugenMayer/docker-image-unison/compare/master...blackjid:docker-image-unison:test_amd) That is branch I was working on, it works. The only issue is that it uses the same job to build both, so it need an arm64 emulator (qemu) to compile the arm64 version, and it's painfully slow..
> 
> Maybe it doesn't matter it is slow because you just compile the image once in a while.
> 
> The way to fix it is to run travis on different jobs each on the correct arch (like the current travis config is doing) and after that create and push the manifest to the registry to bundle both images in one

I'll leave that up to @EugenMayer, I haven't seen any traction there (https://github.com/EugenMayer/docker-image-unison/issues/27) which is why I proposed this solution as this is currently a bug requiring manual intervention for all developers on M1 Macs. 

I do believe the multi architecture image for unison is a better approach, however this solution is better than none. @EugenMayer  Let me know which approach you'd like before I make your proposed suggestions above.

EugenMayer:
@yoseph1998 i really go with what you guys think it's best. I yet did not do a lot of multi arch docker images, but i assume, reading what you write, one has no longer use tags to split different architectures, it is now supported in the registry / by the docker client to pick the right arch for the specific image (if it has been pushed in the first place). If that is correct, i of course would suggest rather fixing this on the unison side and adding proper multiarch images there, instead implementing the logic here.

Thank you for caring and stepping up for it, very appritated
.

yoseph1998:
@EugenMayer I have not worked with the new manifest feature, however it seems like @hufman provided instructions on how to do so here https://github.com/EugenMayer/docker-image-unison/issues/27. It should be straight forward since you already have both the ARM and AMD images compiled.

After that change we'll need to change the version on this repo to not specify the architecture. Once the combined manifest is pushed I can make the accompanying change here. Putting this PR into draft in the meantime

EugenMayer:
considering the effort in https://github.com/EugenMayer/docker-image-unison/pull/31 we should close this issue and superseed it with one that upgrades unison and also no longer uses ARCH suffixes, right?

yoseph1998:
Closing PR in favor of different approach.

PR #815, opened 2022-09-08
Fix sync_prefer dest (sync_host_port not defined)

yoseph1998:
When syncing using Unison and specifying `sync_prefer` as `dest` (line 142/143), the `sync_host_port` variable is not defined.

Fix is similar to how the `sync_host_port` was defined in line 126.

Tested and working locally.

### DIFF
--- a/lib/docker-sync/sync_strategy/unison.rb
+++ b/lib/docker-sync/sync_strategy/unison.rb
@@ -135,6 +135,8 @@ module DockerSync
 
       # cares about conflict resolution
       def sync_prefer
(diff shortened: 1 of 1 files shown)


EugenMayer:
Thank you for looking into that! Is that a regression in 1.0.0? Doesn't seem like that

yoseph1998:
@EugenMayer This isn’t a regression item for 1.0.0, but a defect that seems to have been present for a while but never spotted (not many people would pick ‘dest’ conflict resolution).

EugenMayer:
Thank you a lot for looking at that in detail!

PR #816, opened 2022-09-09
Add default sync excludes (daemon.log & daemon.pid)

yoseph1998:
Add daemon.log and daemon.pid to a default sync excludes directory.

`unison` strategy and `native_osx` strategy changes validated and tested.

**Help wanted to validate rsync strategy.**


EugenMayer (approved):
Nice, this was on the wanted list for some time, thank you. LGTM, also the rsync part looks correct

Issue #819, opened 2022-09-13
Fatal error with Unison because of wrong default exclude definitions


zessx:
### Fatal Error
Exclude definitions from https://github.com/EugenMayer/docker-sync/commit/190d93d44bde47d5c7d9f11e2ef329223bfd1364 are not following the Unison format.

The following error now occurs at each start:
```
success  Unison server started
command  docker inspect --format='{{(index (index .NetworkSettings.Ports "5000/tcp") 0).HostPort}}' wakeo_platform_sync
command  docker inspect --format='{{(index (index .NetworkSettings.Ports "5000/tcp") 0).HostPort}}' wakeo_platform_sync
command  unison -ignore='Path .git' -ignore='Path tmp' -ignore='Path db/backups' -ignore='Path log' -ignore='Path vendor' -ignore='.docker-sync/daemon.log' -ignore='.docker-sync/daemon.pid' -ignore='.docker-sync/daemon.log' -ignore='.docker-sync/daemon.pid' -ignore='.docker-sync/daemon.log' -ignore='.docker-sync/daemon.pid' '/Users/samuel/Projects/wakeo_platform' -auto -batch -prefer '/Users/samuel/Projects/wakeo_platform' -copyonconflict socket://127.0.0.1:51367
  error  Error starting sync, exit code 7
message
message  Fatal error: Usage: unison [options]
        […]

         bad pattern:  .docker-sync/daemon.log
         A pattern must be introduced by one of the following keywords:
             Name, Path, BelowPath or Regex.

success  Starting Docker-Sync in the background
```

When the Unison strategy is used, the `Environment.default_ignores` values must be changed by prepending `Path ` to each of them.

### Docker Driver
No idea (using the native Docker Desktop app for macOS)

### Sync strategy
`unison`

### your docker-sync.yml
```yaml
version: "2"
options:
  verbose: true
syncs:
  wakeo_platform_sync:
    src: .
    sync_excludes_type: none
    sync_excludes:
      - "Path .git"
      - "Path tmp"
      - "Path db/backups"
      - "Path log"
      - "Path vendor"
    sync_strategy: unison 
    sync_prefe: default 
```

### OS
macOS Monterey 12.3

zessx:
After a few tests, it works with the following settings:
```yaml
version: "2"
options:
  verbose: true
syncs:
  wakeo_platform_sync:
    src: .
    sync_excludes_type: Path
    sync_excludes:
      - .git
      - tmp
      - db/backups
[...]

PR #820, opened 2022-09-13
Use the Unison 'Name' prefix for default ignores

zessx:
We cannot rely on `sync_excludes_type` for default ignores, at the user can and will override it for his needs.

Ref: https://github.com/EugenMayer/docker-sync/issues/819


yoseph1998:
@zessx Great catch! Looks like that needs to be modified in the native_osx.rb file as well.

EugenMayer:

PR #821, opened 2022-09-13
Fix nil excludes error (rsync strategy)

yoseph1998:
Fix rsync nil error when sync_excludes is undefined.

### DIFF
--- a/lib/docker-sync/sync_strategy/rsync.rb
+++ b/lib/docker-sync/sync_strategy/rsync.rb
@@ -61,10 +61,9 @@ module DockerSync
 
       def sync_options
         args = []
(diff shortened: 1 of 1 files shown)


vokshirg:
not only one strategy affected https://github.com/EugenMayer/docker-sync/commit/190d93d44bde47d5c7d9f11e2ef329223bfd1364


unison also get: 
```
bad pattern:  .docker-sync/daemon.log
A pattern must be introduced by one of the following keywords:
Name, Path, BelowPath or Regex.
```

EugenMayer:
@yoseph1998 is a dupe of #820

EugenMayer:
Oh, one is for rsync, one is for unison .. and we will probably need one for native_osx too?

EugenMayer:

PR #822, opened 2022-09-14
Replicate usage the Unison 'Name' prefix for default ignores in native_osx strategy

zessx:
Follow-up of this comment https://github.com/EugenMayer/docker-sync/issues/819#issuecomment-1245725820


EugenMayer:
Thank you for fixing the native_osx strategy too!

PR #824, opened 2022-09-15
Fix NameError in DockerSync::SyncStrategy::Rsync#sync_options

mmrko:
Addresses https://github.com/EugenMayer/docker-sync/issues/823.

PR #825, opened 2022-09-15
Fix overriden default ignores

yoseph1998:
During the fix for the exclude type, the default ignores were being overridden by the excludes list.

Seperating them into two and appending + flattening them on returning.


EugenMayer:
Thank you!

EugenMayer:
@yoseph1998 do you think #824 is still needed?

yoseph1998:
@EugenMayer Yes, that's a good catch from mmrko.

rurkss:
you added method `append` with is worked only from ruby 2.5.  No more support for ruby 2.4?

PR #838, opened 2024-07-25
fix(docker-compose): compatibility with docker-compose v2 and docker >= v27

jankal:
As the docker-compose binary was removed and replaced by calling `docker compose` within the `docker` binary as docker-compose v2, we need to update how we call the compose commands here.

See https://docs.docker.com/compose/migrate/

### DIFF
--- a/lib/docker-sync/docker_compose_session.rb
+++ b/lib/docker-sync/docker_compose_session.rb
@@ -43,7 +43,7 @@ module DockerSync
         command_args = ['compose'] + file_args + args
       end
 
-      @last_command = Command.run(command, *docker_args, dir: @dir).join
+      @last_command = Command.run(command, *command_args, dir: @dir).join
       status = @last_command.status
       out = @last_command.captured_output
       err = @last_command.captured_error
--- a/lib/docker-sync/docker_compose_session.rb
+++ b/lib/docker-sync/docker_compose_session.rb
@@ -26,14 +26,24 @@ module DockerSync
 
     private
 
+    def docker_compose_binary_exists?
+      system('which docker-compose > /dev/null 2>&1')
+    end
+
+
     def run!(*args)
       # file_args and args should be Array of String
       file_args = @files.map { |file| "--file=#{file}" }
 
-      # Prepend "compose" to the args array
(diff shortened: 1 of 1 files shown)


EugenMayer:
Thank you for the contribution.

There is an easy hack for this for now

create this in /usr/local/bin/docker-compose
```bash
#!/bin/bash

docker compose --compatibility "$@"
```

Why i kind of like that one, it supports both, the old and new variant. If we switch to your PR, we support dc 2+ only and there is no easy way back.

What is your opinion on that matter?

jankal:
Hi @EugenMayer!

I guess the most correct way to implement this would be to detect if the `docker-compose` binary exists and fall back to `docker compose` if it doesn't - one could also do it the other way round (detect the docker version - default would be `docker compose` - and fall back to `docker-compose` if it doesn't exist).

Unfortunately, my ruby skills are just not enough to implement this efficiently.

I guess `docker-sync` should work with both versions - as a user, I shouldn't have to implement a workaround in my system for a known change.

EugenMayer:
> I guess docker-sync should work with both versions - as a user, I shouldn't have to implement a workaround in my system for a known change.

Sure, as with all convenience features, they still need to be implemented though. 

Checking for a binary too exist with ruby isnt hard, just use some google-fu :)

jankal:
@EugenMayer I just took a shot at it.

Would you review my latest changes?

lib/docker-sync/docker_compose_session.rb
```diff
@@ -26,11 +26,24 @@ def down
 
     private
 
+    def docker_compose_binary_exists?
```
EugenMayer, line 29: we should call this the legacy binary

EugenMayer (changes requested):
minor thing, otherwise LGTM! Thanks

lib/docker-sync/docker_compose_session.rb
jankal: Thx for the suggestion!
Made the change in [19e0ec1](https://github.com/EugenMayer/docker-sync/pull/838/commits/19e0ec1d1dfeb6caed13f516a1ce65e47fa8482e)

Issue #833, opened 2023-02-11
docker-sync start failed - /tmp/supervisor.sock no such file


digidycom:
# Problem
When I start `docker-sync` for the first time after I run docker, I got following error:

```
docker-sync start
          ok  Starting native_osx for sync my-project-sync
          ok  my-project-sync container not running
          ok  starting my-project-sync container
     command  docker start my-project-sync && docker exec my-project-sync supervisorctl restart unison
my-project-sync
unix:///tmp/supervisor.sock no such file
/Users/martin/.local/share/gem/ruby/2.6.0/gems/docker-sync-1.0.5/lib/docker-sync/sync_strategy/native_osx.rb:105:in `start_container': Start failed (RuntimeError)
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/docker-sync-1.0.5/lib/docker-sync/sync_process.rb:105:in `start_container'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/docker-sync-1.0.5/lib/docker-sync/sync_manager.rb:103:in `block in start_container'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/docker-sync-1.0.5/lib/docker-sync/sync_manager.rb:102:in `each'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/docker-sync-1.0.5/lib/docker-sync/sync_manager.rb:102:in `start_container'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/docker-sync-1.0.5/tasks/sync/sync.thor:181:in `daemonize'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/docker-sync-1.0.5/tasks/sync/sync.thor:47:in `start'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/thor-1.2.1/lib/thor/command.rb:27:in `run'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/thor-1.2.1/lib/thor/invocation.rb:127:in `invoke_command'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/thor-1.2.1/lib/thor.rb:392:in `dispatch'
	from /Users/martin/.local/share/gem/ruby/2.6.0/gems/thor-1.2.1/lib/thor/base.rb:485:in `start'
[...]

PR #842, opened 2024-10-22
Fixing docker-sync start fail on docker engine (re)start (issue #833)

igronus:
The apoach with hunting for `/tmp/supervisor.sock` won't work (tried that as well with many variations) because it's apparently using by internal docker mechansim while starting the `supervisorctl` service: seems like the file is temporary created only not to run two start commands for the service, as it is common practice in POSIX systems. After the service is responding, the file doesn't exist anymore.


EugenMayer (approved):
Thank you for the contibution! looks fine to me

